PR #14, opened 2020-03-06
Integer plus

PetarMax:
Integer reasoning works on WISL example


giltho:
Closing, #16 contained this

PR #24, opened 2020-03-26
Add import mechanism to Gillian-C

alexismarinoiu:


Gillian-C/lib/parserAndCompiler.ml
```diff
@@ -81,10 +85,80 @@ let parse_and_compile_file path =
       close_out oc
   in
   let annots = parse_annots path in
+  Gilgen.trans_program_with_annots exec_mode last_clight csm annots
+
+module Symbol_set = Set.Make (String)
```
giltho: We already instantiated StringSets as `Gillian.Utils.Containers.SS` :)
```suggestion
module Symbol_set = Gillian.Utils.Containers.SS
```

GillianCore/GIL_Syntax/Literal.ml
```diff
@@ -26,7 +31,7 @@ let rec pp fmt x =
   | Constant c -> Fmt.string fmt (Constant.str c)
   | Bool b     -> if b then Fmt.string fmt "true" else Fmt.string fmt "false"
   | Int i      -> Fmt.int fmt i
-  | Num n      -> Fmt.float fmt n
+  | Num n      -> Fmt.pf fmt "%s" (remove_trailing_dot (Printf.sprintf "%F" n))
```
giltho, line 34: Wait I don't get that, why did you remove the trailing dot ? We're about to add proper integers to gillian, and when we do, we need the trailing dot to differentiate between floats and integers
Also, right now, the regexp for floats is
```ocaml
let float = '-'? digit+ ('.' digit*)?
```
which is compatible with a trailing dot
Sorry, I think you explained earlier but I didn't get it

alexismarinoiu: Okay, I thought it was a bit redundant when there are no digits after the decimal points (or only zeros). I'll change it to use `%F`.

PR #22, opened 2020-03-25
[WIP] Better reporting system

raduml97:
Interface of reports for #21

PR #28, opened 2020-04-06
Separate logging from utils

raduml97:
Since #22 plans to expand `logging` into a large module, separating it from `utils` into its own module seems the logical approach.

Signed-off-by: Radu Lacraru <radu.lacraru97@gmail.com>


giltho:
merging as soon as tests pass

PR #33, opened 2020-04-22
Allow multi-file analysis of JS programs

alexismarinoiu:
This involves calling the updated `JS_Parser` interface in the correct way and rewriting the Buckets.js tests to have the library in a single file and have each test import it using `require`.


giltho:
I don't know why the checks aren't printed correctly but they all passed

PR #34, opened 2020-04-23
Fix slowness when running bulk tests in Gillian-C

alexismarinoiu:
This involves caching the parsed and compiled files so that they can be reused across different test runs but within the same running instance of Gillian-C.

### DIFF
--- a/GillianCore/CommandLine/CommandLine.ml
+++ b/GillianCore/CommandLine/CommandLine.ml
@@ -98,8 +98,8 @@ struct
 
   let no_print_failures =
     let doc =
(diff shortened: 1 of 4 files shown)


GillianCore/CommandLine/CommandLine.ml
```diff
@@ -98,8 +98,8 @@ struct
 
   let no_print_failures =
     let doc =
-      "Do not print the list of all the failed tests at the end of execution \
-       bulk execution"
+      "Do not print the list of all the failed tests at the end of the bulk \
+       execution"
```
giltho, line 102: 😆

PR #39, opened 2020-05-04
Add support for continuous verification

alexismarinoiu:
This is intended to move Gillian towards being able to perform incremental analysis, starting with verification. It adds a new command-line mode, `--incremental`, which, if enabled, causes it to work out the minimal number of procedures that need to be re-verified based on the information it derives about what has changed. It makes use of a results directory (currently named `.gillian`) which sits in the current directory where Gillian is invoked.


GillianCore/engine/Abstraction/VerificationResults.ml
```diff
@@ -0,0 +1,45 @@
...
+
+let merge_results results other_results =
+  let () =
+    Hashtbl.iter
+      (fun id verified -> Hashtbl.replace results id verified)
+      other_results
+  in
+  results
+
+let to_json results =
```
giltho, line 19: What about using `ppx_conv_yojson` instead ?
Maybe in staged mode mode to make it clearer

GillianCore/engine/Abstraction/VerificationResults.ml
```diff
@@ -0,0 +1,45 @@
+module Json = Yojson.Basic
```
giltho, line 1: Yojson.Safe is a better alternative, because it's supported by the derivers

Gillian-C/lib/parserAndCompiler.ml
```diff
@@ -107,9 +109,9 @@ module TargetLangOptions = struct
     Config.hide_mult_def := hide_mult_def
 end
 
-let compiled_progs = Hashtbl.create 16
+let compiled_progs = Hashtbl.create small_tbl_size
 
-let gil_paths = Hashtbl.create 16
+let gil_paths = Hashtbl.create small_tbl_size
```
giltho, line 114: Do I have a way of resetting these tables ? 
I need it if I want to execute several programs without restarting the program

GillianCore/CommandLine/CommandLine.ml
```diff
@@ -405,12 +413,32 @@ struct
...
+          let results = VerificationResults.merge_results results cur_results in
+          write_results_dir { sources = cur_source_paths; call_graph; results }
+        else
+          (* Analyse all procedures *)
+          let to_verify = Containers.to_key_set prog.procs in
+          let results, call_graph =
+            Verification.verify_procs prog to_verify ()
+          in
+          write_results_dir { sources = cur_source_paths; call_graph; results }
+      in
```
giltho, line 441: That should maybe be defined in Verification directly ?

GillianCore/engine/Abstraction/VerificationResults.ml
giltho, line 19: This should be named "to_yojson" and you should also use "of_yojson" everywhere
What about using ppx_yojson_conv ?

GillianCore/results/resultsDir.ml
```diff
@@ -0,0 +1,82 @@
+open Containers
+module Json = Yojson.Basic
```
giltho, line 2: Yojson.Safe is a better alternative as it is compatible with the ppx derivers

GillianCore/engine/Abstraction/VerificationResults.ml
giltho: In stage mode so that the code is accessible

GillianCore/results/resultsDir.ml
```diff
@@ -0,0 +1,82 @@
...
+  {
+    sources = SourcePaths.from_json (read_json Filenames.sources);
+    call_graph = CallGraph.from_json (read_json Filenames.call_graph);
+    results = VerificationResults.from_json (read_json Filenames.verif_results);
+  }
+
+let write_results_dir { sources; call_graph; results } =
+  let write_json json filename =
+    let json_path = Filename.concat results_dir filename in
+    Json.to_file ~std:true json_path json
```
giltho, line 76: Biniou should maybe be a choice instead of Json. You don't have to do it now, just take not of that somewhere and either you'll do it later or I'll do it whenver

GillianCore/GIL_Parser/GIL_Lexer.mll
```diff
@@ -262,8 +262,11 @@ rule read = parse
   | loc                  { GIL_Parser.LOC (Lexing.lexeme lexbuf) }
   | aloc                 { GIL_Parser.ALOC (Lexing.lexeme lexbuf) }
   | normalised_aloc      { GIL_Parser.ALOC (Lexing.lexeme lexbuf) }
-
-  (* Variables: THIS IS NEW *)
+(* Directives *)
+  | "@nopath"            { GIL_Parser.NO_PATH }
+  | "@internal"          { GIL_Parser.INTERNAL }
+  | "internal;"          { GIL_Parser.INTERNAL_FILE }
```
giltho, line 268: Minor, but can we go with
`@@internal` or `#internal` and drop the `;` at the end ?
I slightly prefer the second one because the syntax already exists in the wild (C)

GillianCore/GIL_Parser/parser_state.ml
```diff
@@ -0,0 +1,7 @@
+let procs_with_no_paths = ref Containers.SS.empty
+
+let internal_file = ref false
```
giltho, line 3: Are you sure you want to make this a global reference ?
`procs_with_no_paths` is ok because it's program-wide, but `internal_file` is proc-wide, it's gonna change very often. I would prefer if you create a state record type that you'd pass to every function (using a tagged parameter)
Actually in a better world it we'd have a compilation-state monad but I'm not gonna make you do that.

Changing this is optional though, but at least write a `FIXME` next to it explaining, because it will not be compatible with multi-core ocaml when it becomes usable

GillianCore/engine/Abstraction/Verifier.ml
```diff
@@ -453,6 +484,40 @@ struct
...
         L.normal (fun m -> m "%s" msg)
+
+  let prune_call_graph call_graph proc_names =
+    let proc_ids = List.map CallGraph.id_of_proc_name proc_names in
+    List.iter (CallGraph.remove call_graph) proc_ids
+
+  let prune_results results proc_names =
+    List.iter (VerificationResults.remove results) proc_names
+
+  let verify_prog (prog : (Annot.t, int) Prog.t) (incremental : bool) : unit =
```
giltho, line 495: I'll take a much closer look at this, you'll walk me through it step by step on zoom 😅

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
```diff
@@ -127,6 +131,11 @@ struct
            ------------------------------------------------------@]@\n"
           (Sys.time ()) LCmd.pp lcmd State.pp state)
 
+  let is_internal_proc (prog : UP.prog) proc_name =
+    match Prog.get_proc prog.prog proc_name with
+    | Some proc -> proc.proc_internal
+    | None      -> false
```
giltho, line 137: Can this case ever happen ? If you think not, use `failwith` so we can detect inconsistent cases when they happen

wisl/lib/ParserAndCompiler/wisl2Gil.mli
```diff
@@ -1,3 +1,3 @@
 open Gillian.Gil_syntax
 
-val compile : WProg.t -> (Annot.t, string) Prog.t
+val compile : string -> WProg.t -> (Annot.t, string) Prog.t
```
giltho, line 3: can you make that `compile : filename:string -> WProg.t -> (Annot.t, string) Prog.t` ?
I like to use tagged parameters when it has a common type like  `string` and it's impossible to know what it's used for from the signature. Here, the string parameter is very non-informative :/

If you can make the same change for C that would be amazing

GillianCore/GIL_Syntax/Gil_syntax.mli
```diff
@@ -872,24 +872,36 @@ module Prog : sig
...
+
+  val get_noninternal_lemma_names : ('a, 'b) t -> string list
+  (** Get names of all lemmas not marked as internal *)
+
   val get_proc : ('a, 'b) t -> string -> ('a, 'b) Proc.t option
   (** Get a specific procedure *)
 
   val get_proc_exn : ('a, 'b) t -> string -> ('a, 'b) Proc.t
-  (** Get a specific procedure *)
+  (** Get a specific procedure. Raises [Failure] if it does not exist *)
```
giltho, line 885:
```suggestion
  (** Get a specific procedure. Raises [Failure] if it does not exist *)
  (* FIXME: should raise Not_found instead *)
```

PR #40, opened 2020-05-06
[WIP] Logging: language dependent logging

raduml97:
Allows for different target languages to log different types of content.

### DIFF
--- a/GillianCore/logging/report.ml
+++ b/GillianCore/logging/report.ml
@@ -6,18 +6,19 @@ let string_of_phase = function
   | Preprocessing       -> "Preprocessing"
   | Verification        -> "Verification"
 
-type 'a content =
-  | Debug of ((('a, Format.formatter, unit) format -> 'a) -> unit)
-  | Phase of phase
+type ('a, 'b) content =
+  | Debug      of ((('a, Format.formatter, unit) format -> 'a) -> unit)
+  | Phase      of phase
+  | TargetLang of 'b
 
 type severity = Info | Log | Success | Error | Warning
 
-type 'a t = {
+type ('a, 'b) t = {
   id : Uuidm.t;
   title : string;
   elapsed_time : float;
   previous : Uuidm.t option;
   parent : Uuidm.t option;
-  content : 'a content;
+  content : ('a, 'b) content;
   severity : severity;
 }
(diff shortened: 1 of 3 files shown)


Gillian-C/lib/External.ml
```diff
@@ -1,5 +1,6 @@
 open Gillian.Gil_syntax
 open Gillian.General
+module L = Gillian.Logging.Default
```
giltho, line 3: Not liking this, if you change the logger you change everywhere
Maybe create a Logging module in `Gillian-C` that just contains `include Gillian.Logging.Default`

giltho: Then you can also just call `Logging` everywhere, which is nicer too

GillianCore/logging/logging.ml
```diff
@@ -1,129 +1,250 @@
...
-let log lvl msgf =
-  if Mode.should_log lvl then
-    let report = ReportBuilder.info "" (Debug msgf) () in
-    Reporter.log report
+  let reporters =
+    [ new Reporter.file_reporter (); new Reporter.database_reporter () ]
+end
 
-let normal msgf = log Normal msgf
+module type S = sig
```
giltho, line 11: One solution you could adopt to reduce duplication is creating a `reporter_S.mli` which just contains the signature.
In the dune file you will need to add `(modules_without_implementation reporter_S)`
Then you should be able to say things like
```ocaml
module Make(P: TargetLang.S) : Reporter_S
module Default : Reporter_S
```

I find `Reporter_S` a little less elegant than `Reporter.S`, but code duplication is difficult to maintain

giltho:
Ok, I think there's a tiny thing wrong with the design we've chosen
If I want to report something during the execution of an action, I need my logger instantiated. But to instantiate the logger I need the types of my memory model to be defined and I need to have instantiated my memory model

One solution is probably passing the reporters to the actions.

My second issue is, you're importing `Reporter.Default` everywhere in the code.
If I change my reporter module, do I need to make every module in Gillian a functor parametrized by the Reporter module ? That can't work. There has to be a better solution. I need to think about it.

raduml97:
Will reopen a new PR for this.

PR #46, opened 2020-05-14
Logging: language dependent logging

raduml97:
Based on #45, will rebase after it gets merged.

### DIFF
--- a/GillianCore/logging/databaseReporter.ml
+++ b/GillianCore/logging/databaseReporter.ml
@@ -76,3 +76,11 @@ class virtual ['a] t =
 
     method wrap_up = wrap_up ()
   end
+
+let default : type a. unit -> a t =
+ fun () ->
+  object
+    inherit [a] t
+
+    method private serialize_target_lang _ = ""
+  end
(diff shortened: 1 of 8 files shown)


wisl/lib/semantics/wislCMemory.ml
```diff
@@ -11,6 +11,32 @@ type fix_t = unit
...
+      method private serialize_target_lang = WislCHeap.str
+    end
+
+  let reporters =
+    [
+      (file_reporter :> t Logging.Reporter.t);
+      (database_reporter :> t Logging.Reporter.t);
+    ]
+end)
+
```
raduml97, line 39: For demonstrative purposes, will remove before merging.

giltho: Ok sorry I should have reviewed that yesterday, but you have a lot of things to do today anyway
If I may add something, we need to bite a bullet here that I've tried to avoid but it seems necessary before merging this.
Can you write atd generators for the wisl memory type and serialize it into memory ?

PR #48, opened 2020-05-18
Add `cmitomli` as a dev dependency

raduml97:
Signed-off-by: Radu Lacraru <radu.lacraru97@gmail.com>


.ocamlformat
```diff
@@ -1,3 +1,4 @@
+version=0.14.2
```
raduml97, line 1: Nice one

PR #53, opened 2020-05-22
Logging: database logging using JSON

raduml97:
Database logging no longer uses `sanddb`, rather, it outputs an array of JSON objects each representing a report.


GillianCore/logging/report.ml
```diff
@@ -1,49 +1,63 @@
+type uuidm = Uuidm.t
+
+let yojson_of_uuidm uuidm = Uuidm.to_string uuidm |> yojson_of_string
+
+let uuidm_of_yojson yojson =
+  string_of_yojson yojson |> Uuidm.of_string |> Option.get
+
 module PackedPP : sig
-  type t
+  type t [@@deriving yojson]
```
giltho, line 9: The deriving yojson should probably go away ?

raduml97: If we remove it, then `[@@deriving yojson]` for `type agnostic_content` will complain because it cannot find `PackedPP.yojson_of_t` and `PackedPP.t_of_yojson` in the signature of `PackedPP`. I guess those could be added in manually as well.

giltho: image
My bad

PR #55, opened 2020-05-22
Add support for continuous symbolic testing

alexismarinoiu:
This is intended to add support for incremental symbolic testing, where Gillian only re-runs the tests that are directly affected by changes made by the user.

### DIFF
--- a/Gillian-C/scripts/testSymb.sh
+++ b/Gillian-C/scripts/testSymb.sh
@@ -8,6 +8,7 @@ if [[ -z "${GITHUB_ACTIONS}" ]]; then
   source exec.env
 fi
 
+# TODO (Alexis): Make incremental analysis thread-safe to allow the use of --parallel
 echo "--- testing SLL ---"
 time gillian-c wpst symbolic/sll.c -l disabled
 rc=$?; if [[ $rc != 0 ]]; then FINAL_RETURN=1; fi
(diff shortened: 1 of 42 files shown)


GillianCore/CommandLine/CommandLine.ml
```diff
@@ -321,26 +321,68 @@ struct
   end
 
   module SInterpreterConsole = struct
-    let process_files files already_compiled outfile_opt =
-      let e_prog =
-        if not already_compiled then (
+    let run (prog : UP.prog) incremental source_files =
```
giltho, line 324: Not a 100% sure this should be defined here, but I'll just change it later

Gillian-C/scripts/testSymb.sh
```diff
@@ -8,33 +8,34 @@ if [[ -z "${GITHUB_ACTIONS}" ]]; then
   source exec.env
 fi
 
+# TODO (Alexis): Make incremental analysis thread-safe to allow the use of --parallel
 echo "--- testing SLL ---"
-time gillian-c wpst symbolic/sll.c -l disabled --parallel
+time gillian-c wpst symbolic/sll.c -l disabled
```
giltho, line 13: I don't know if the --parallel should be removed in the testSymb, this file is not for incremental use ?

PR #59, opened 2020-05-26
adding a customisable unroll and spec re-use mechanism in bi-abduction

PetarMax:


giltho:
I'm afraid this might be incompatible with @alexismarinoiu's work.
He has a decision procedure for what to reanalyse, is this procedure still correct ? I'm thinking it might be correct but over-approximating, which is ok, but I want to make sure

PR #64, opened 2020-06-12
Fix the `loop.wisl` example

raduml97:


giltho:
Why was this closed ?

raduml97:
@giltho build fails for some reason

PR #65, opened 2020-06-12
Fix pointer to scalars in bispecs

giltho:
For @raduml97 
Can you check that, given test.c, specs.expected is what you want ?


raduml97:
> For @raduml97
> Can you check that, given test.c, specs.expected is what you want ?

Yes, it works as expected now

giltho:
I think there was an update in GCC that's breaking our -D thing

PR #103, opened 2022-05-23
Get the debugger working like it used to

NatKarmios:
This is in preparation for unification additions.

A notable change is that GInterpeter step functions now return a branch_case, which is necessary for working with the new log structure when debugging.


GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
```diff
@@ -910,12 +939,26 @@ struct
...
     (* Action *)
     | LAction (x, a, es) -> (
+        Debugger_log.to_rpc
+          ~json:
+            [
+              ("x", `String x);
+              ("a", `String a);
+              ("es", `List (List.map Expr.to_yojson es));
+            ]
+          "LAction";
```
giltho: So, by doing this, you're explicitely building some yojsons.
The whole goal of the logging in place is to have as little overhead as possible when in `-l disabled` mode, that is why every log call is captured in a `(fun fmt -> ...)`, the thing inside the `...` is not computed ever if in `-l disabled`.
Could you try to do something similar here?

GillianCore/logging/databaseReporter.ml
```diff
@@ -6,7 +6,8 @@ let initialize () =
   LogDatabase.reset_db ();
   LogDatabase.create_db ()
 
-let will_log (type_ : string) = type_ <> LoggingConstants.ContentType.debug
+let will_log (type_ : string) =
+  not (List.mem type_ LoggingConstants.ContentType.[ debug; phase ])
```
giltho, line 10: same, try avoiding to build lists for no reason
Also, linked list is the de-facto data structure in ocaml, but it's slower than arrays (cache, pointer indirection etc...)
So in that case I would even prefer
```
not (Array.mem type_ [| debug; phase |])
```
but in any case I would prefer neither of those :p

GillianCore/logging/logDatabase.ml
```diff
@@ -114,17 +114,13 @@ let get_report id =
 
 let get_previous_report_id id =
   let db = get_db () in
-  let stmt =
-    Sqlite3.prepare db
-      "SELECT id FROM report WHERE elapsed_time < (SELECT elapsed_time FROM \
-       report WHERE id=?) AND type='cmd_step' ORDER BY elapsed_time DESC LIMIT \
-       1;"
-  in
```
giltho, line 122: Nice

giltho (dismissed):
Some tiny changes, but overall great work as always

GillianCore/debugging/debugger/debugger.ml
```diff
@@ -51,8 +54,15 @@ struct
...
     mutable errors : err_t list;
     mutable cur_cmd : (int Cmd.t * Annot.t) option;
+    mutable branch_case : branch_case option;
   }
 
+  type branch_case_option = Case of branch_case | TakeFirst | NoCase
+
+  let to_option = function
+    | Some case -> Case case
+    | None -> NoCase
```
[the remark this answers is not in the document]
NatKarmios: Resolved in 13bb507b

GillianCore/debugging/debugger/debugger.ml
```diff
@@ -227,22 +237,79 @@ struct
...
+                    ("results", `List results_json);
+                  ]
+                in
+                raise (Log.FailureJson (msg, json))
+            | Some (id, _) ->
+                update_report_id_and_inspection_fields id NoCase dbg)
+        | t
+          when List.mem t
+                 L.LoggingConstants.ContentType.
+                   [ cmd_step; cmd_result; proc_init ] -> (
```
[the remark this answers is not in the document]
NatKarmios: Resolved in e2b58255

GillianCore/logging/databaseReporter.ml
NatKarmios: Resolved in e2b58255

GillianCore/debugging/debugger/debugger.ml
```diff
@@ -323,23 +391,30 @@ struct
...
     match dbg.cont_func with
-    | None -> ReachedEnd
+    | None ->
+        "No cont_func; reached end" |> Log.to_rpc;
+        ReachedEnd
     | Some cont_func -> (
-        let cont_func = cont_func () in
-        match cont_func with
+        let cont_func_result = cont_func () in
+        match cont_func_result with
```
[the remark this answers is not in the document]
NatKarmios: Resolved in e2b58255

giltho:
PS: your version of debugger.log is so much cleaner, thanks

GillianCore/debugging/log/Debugger_log.ml
```diff
@@ -0,0 +1,74 @@
...
+  match !rpc_ref with
+  | Some rpc ->
+      (info
+         (match json with
+         | _ :: _ -> msg ^ " (+)"
+         | [] -> msg);
+       Debug_rpc.send_event rpc (module Log_event) { msg; json })
+      |> ignore
+  | None -> ()
+
```
giltho, line 50: ~You always use `to_rpc` by passing a `~json` argument, so I don't know if it should be optional~

GillianCore/debugging/log/Debugger_log.ml
```diff
@@ -0,0 +1,74 @@
...
+              ("type", `String type_);
+            ]
+      | None -> `Null
+    in
+    to_rpc ~json:[ ("report", report_json) ] msg
+
+let log f =
+  if enabled () then
+    let msg, json = f () in
+    to_rpc ~json msg
```
giltho, line 68: OCaml's pretty good and fast with continuations (because of some tail optims)
But what that means is probably that you should be doing that instead
```suggestion
let log f =
  if enabled () then f to_rpc
```

And then calling it by doing
```
DL.log (fun t -> t  ~json:blabla msg)
```

aaanndd now I understand why ~json was optional

GillianCore/debugging/log/Debugger_log.ml
giltho: Scratch that, because of next comment

giltho (dismissed):
Just change that detail, and ready to merge!

giltho (approved):

PR #108, opened 2022-07-22
symbol() is now used to generate fresh svars

giltho:
- create a FreshSvar LCmd to create symbolic variables, usable in a loop
- Removes the SpecVar command, which was only used to create fresh svars anyway
- Change the signature of AssumeType, which now takes an expression. It always could have anyway, since it's reverse-typing the expression anyway

### DIFF
--- a/Gillian-C/runtime/binops_common.gil
+++ b/Gillian-C/runtime/binops_common.gil
@@ -165,7 +165,7 @@ proc i__binop_mod(v1, v2) {
 proc i__binop_mulfs(v1, v2) {
         goto [ l-nth(v1, 0i) = "single" ] llon unde;
   llon: goto [ l-nth(v2, 0i) = "single" ] blon unde;
-  blon: ret := {{ "single", l-nth(v1, 1i) i* l-nth(v2, 1i) }};
+  blon: ret := {{ "single", l-nth(v1, 1i) * l-nth(v2, 1i) }};
         return;
   unde: fail[operator]("Using *s operator for non-long elements")
 };
--- a/Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
+++ b/Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
@@ -2198,7 +2198,7 @@ let rec translate_expr tr_ctx e :
       let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
       let x_v = PVar x_v in
       let cmd2 =
-        (metadata, None, LLogic (LCmd.AssumeType (x_v, Type.StringType)))
+        (metadata, None, LLogic (LCmd.AssumeType (x_v, Type.BooleanType)))
       in
       ([ cmd1; cmd2 ], x_v, [])
   | JS_Parser.Syntax.Call (e_f, xes)
--- a/Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
+++ b/Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
   | JS_Parser.Syntax.Call (e_f, xes)
     when e_f.JS_Parser.Syntax.exp_stx
          = JS_Parser.Syntax.Var js_symbolic_constructs.js_symb ->
-      let x =
-        match List.map (fun xe -> xe.JS_Parser.Syntax.exp_stx) xes with
-        | [ JS_Parser.Syntax.Var x ] -> "#" ^ x
+      let () =
+        match xes with
+        | [] -> ()
         | _ -> raise (Failure "Invalid symbolic")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
+      let cmd1 = (metadata, None, LLogic (LCmd.FreshSVar x_v)) in
+      let x_v = PVar x_v in
+      let cmd2 =
+        (metadata, None, LLogic (LCmd.Assume (Not (Eq (x_v, Lit Empty)))))
+      in
       let cmd3 =
-        (metadata, None, LLogic (LCmd.Assume (Not (Eq (LVar x, Lit Empty)))))
+        (metadata, None, LLogic (LCmd.Assume (Not (Eq (x_v, Lit Nono)))))
       in
       let cmd4 =
-        (metadata, None, LLogic (LCmd.Assume (Not (Eq (LVar x, Lit Nono)))))
-      in
-      let cmd5 =
         ( metadata,
           None,
           LLogic
-            (LCmd.Assume (Not (Eq (UnOp (TypeOf, LVar x), Lit (Type ListType)))))
+            (LCmd.Assume (Not (Eq (UnOp (TypeOf, x_v), Lit (Type ListType)))))
         )
       in
-      ([ cmd1; cmd2; cmd3; cmd4; cmd5 ], PVar x_v, [])
+      ([ cmd1; cmd2; cmd3; cmd4 ], x_v, [])
   | JS_Parser.Syntax.Call (e_f, xes)
     when e_f.JS_Parser.Syntax.exp_stx
          = JS_Parser.Syntax.Var js_symbolic_constructs.js_symb_number ->
-      let x =
-        match List.map (fun xe -> xe.JS_Parser.Syntax.exp_stx) xes with
-        | [ JS_Parser.Syntax.Var x ] -> "#" ^ x
+      let () =
+        match xes with
+        | [] -> ()
         | _ -> raise (Failure "Invalid symb_number")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
-        (metadata, None, LLogic (LCmd.AssumeType (x, Type.NumberType)))
+      let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
+      let x_v = PVar x_v in
+      let cmd2 =
+        (metadata, None, LLogic (LCmd.AssumeType (x_v, Type.NumberType)))
       in
-      ([ cmd1; cmd2; cmd3 ], PVar x_v, [])
+      ([ cmd1; cmd2 ], x_v, [])
   | JS_Parser.Syntax.Call (e_f, xes)
     when e_f.JS_Parser.Syntax.exp_stx
          = JS_Parser.Syntax.Var js_symbolic_constructs.js_symb_string ->
-      let x =
-        match List.map (fun xe -> xe.JS_Parser.Syntax.exp_stx) xes with
-        | [ JS_Parser.Syntax.Var x ] -> "#" ^ x
-        | _ -> raise (Failure "Invalid symb_number")
+      let () =
+        match xes with
+        | [] -> ()
+        | _ -> raise (Failure "Invalid symb_string")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
-        (metadata, None, LLogic (LCmd.AssumeType (x, Type.StringType)))
+      let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
+      let x_v = PVar x_v in
+      let cmd2 =
+        (metadata, None, LLogic (LCmd.AssumeType (x_v, Type.StringType)))
       in
-      ([ cmd1; cmd2; cmd3 ], PVar x_v, [])
+      ([ cmd1; cmd2 ], x_v, [])
   | JS_Parser.Syntax.Call (e_f, xes)
     when e_f.JS_Parser.Syntax.exp_stx
          = JS_Parser.Syntax.Var js_symbolic_constructs.js_symb_bool ->
-      let x =
-        match List.map (fun xe -> xe.JS_Parser.Syntax.exp_stx) xes with
-        | [ JS_Parser.Syntax.Var x ] -> "#" ^ x
+      let () =
+        match xes with
+        | [] -> ()
         | _ -> raise (Failure "Invalid symb_bool")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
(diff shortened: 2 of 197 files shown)


GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
```diff
@@ -541,25 +541,19 @@ struct
 
     let eval_expr = make_eval_expr state in
     match lcmd with
-    | AssumeType (x, t) -> (
-        match Val.from_expr (LVar x) with
+    | AssumeType (e, t) -> (
+        match Val.from_expr e with
```
PetarMax, line 545: Should this not evaluate the expression properly?

GillianCore/utils/generators.ml
```diff
@@ -28,8 +28,11 @@ let fresh_lvar, reset_lvar = fresh_sth lvar_prefix
 (** Logical variable counter *)
 let fresh_lvar_bi, reset_lvar_bi = fresh_sth lvar_prefix_bi
 
+let fresh_svar, reset_svar = fresh_sth "#gen__"
```
PetarMax, line 31: How does this prefix interact with, say, `#var` or just `#` later on?

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
```diff
@@ -587,7 +581,11 @@ struct
                | Some state' -> [ state' ]
                | _ -> [])
              fos)
-    | SpecVar xs -> [ State.add_spec_vars state (Var.Set.of_list xs) ]
+    | FreshSVar x ->
+        let new_svar = Generators.fresh_svar () in
+        let state' = State.add_spec_vars state (SS.singleton x) in
+        let v = State.eval_expr state' (LVar new_svar) in
```
PetarMax, line 587: Ah, so here the expression is evauated properly (see below)...

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
```diff
@@ -587,7 +581,11 @@ struct
                | Some state' -> [ state' ]
                | _ -> [])
              fos)
-    | SpecVar xs -> [ State.add_spec_vars state (Var.Set.of_list xs) ]
+    | FreshSVar x ->
+        let new_svar = Generators.fresh_svar () in
+        let state' = State.add_spec_vars state (SS.singleton x) in
```
PetarMax, line 586: This seems to be a program variable `x`? If so, it can't go into the `spec_vars`, `v` should go into the spec vars?

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
```diff
@@ -587,7 +581,11 @@ struct
...
+        let v = State.eval_expr state' (LVar new_svar) in
+        [ update_store state' x v ]
```
PetarMax, line 588: Should we check that `x` actually has the format of a program variable name?

Gillian-JS/lib/Parsing/Javert_Lexer.mll
```diff
@@ -122,7 +122,7 @@
       "assert",       Javert_Parser.ASSERT;
       "assume",       Javert_Parser.ASSUME;
       "assume_type",  Javert_Parser.ASSUME_TYPE;
-      "spec_var",     Javert_Parser.SPEC_VAR;
+      "symbol", Javert_Parser.SYMBOL;
```
PetarMax, line 125: Alignment?

GillianCore/CommandLine/CommandLine.ml
```diff
@@ -447,6 +447,7 @@ struct
         incremental
         entry_point
         () =
+      let () = Fmt_tty.setup_std_outputs () in
```
PetarMax, line 450: I think this is separate from the spec vars?

Gillian-C/runtime/binops_common.gil
```diff
@@ -165,7 +165,7 @@ proc i__binop_mod(v1, v2) {
 proc i__binop_mulfs(v1, v2) {
         goto [ l-nth(v1, 0i) = "single" ] llon unde;
   llon: goto [ l-nth(v2, 0i) = "single" ] blon unde;
-  blon: ret := {{ "single", l-nth(v1, 1i) i* l-nth(v2, 1i) }};
+  blon: ret := {{ "single", l-nth(v1, 1i) * l-nth(v2, 1i) }};
```
PetarMax, line 168: Not sure what happened here?

GillianCore/GIL_Parser/GIL_Lexer.mll
```diff
@@ -110,7 +110,7 @@
       "assert",       GIL_Parser.ASSERT;
       "assume",       GIL_Parser.ASSUME;
       "assume_type",  GIL_Parser.ASSUME_TYPE;
-      "spec_var",     GIL_Parser.SPEC_VAR;
+      "symbol",       GIL_Parser.SYMBOL;
```
PetarMax, line 113: It's a symbolic variable - I'm more for sth along the lines of `fresh_svar`.

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
```diff
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
...
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
-        (metadata, None, LLogic (LCmd.AssumeType (x, Type.BooleanType)))
+      let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
+      let x_v = PVar x_v in
+      let cmd2 =
+        (metadata, None, LLogic (LCmd.AssumeType (x_v, Type.StringType)))
```
PetarMax, line 2201: `BooleanType`!

PetarMax (changes requested):
Some comments...

GillianCore/utils/generators.ml
giltho: I tried to choose a prefix that wasn't used anywhere else, but that could be broken by a future instantiation
I believe we should intern all variable names, and that would make it easier to ensure freshness when generating them
But even if it can be broken in the future, this is still more resilient than what was there before

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: So, I agree, but Jose's code was saying that he needs to evaluate it and I don't know why 🙃

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: Oh god you're right!

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
PetarMax: I mean, yes, we can, but in this case the two are equivalent - we should choose one, probably expression evaluation rather than `Val.from_expr`.

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: Probably yes. Ideally, we wouldn't use strings for variables and it would be type-sound...

GillianCore/utils/names.ml
```diff
@@ -19,6 +19,8 @@ let aloc_prefix = "_$l_"
 (** Logical variable prefix  *)
 let lvar_prefix = "_lvar_"
 
+let svar_prefix = "#var_"
```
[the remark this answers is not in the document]
giltho: Ah wait, this line should simply go away!

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: I'm not sure it should, the goal is to reverse-type the expression itself, and then add whatever types to the typenv, it's not really required to evaluate the expression, and it's a logic command, so Val.from_expr is guaranteed to exist

giltho: Ah, but I do want to evaluate it, because I want to type the lvar properly, not the pvar

Gillian-C/runtime/binops_common.gil
giltho: I fixed also the bulk execution in Gillian-C... Some tests were dying actually, and I wasn't detecting that
That was one reason

GillianCore/GIL_Parser/GIL_Lexer.mll
giltho: fresh_svar it is

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
```diff
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
   | JS_Parser.Syntax.Call (e_f, xes)
```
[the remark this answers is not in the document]
giltho: I know why it was there: it was giving a name to the variable.
var x = symb_num(y) -> x : #y in the store
But now that we're generating them dynamically, that doesn't work anymore.

GillianCore/utils/generators.ml
giltho: Got it now, #var just doesn't exist

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
giltho: Fixed

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: FIxed

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: I decided, it's going to be Val.from_expr. It's cleaner in theory: the lvar is fresh, it always evaluates to itself, and any other interpretation would be wrong.

GillianCore/engine/GeneralSemantics/General/GInterpreter.ml
giltho: It's not done in Assignment, so I'm just not going to do it there.

GillianCore/CommandLine/CommandLine.ml
giltho: That is because, plot twist, failure during wpst would not result in any failure at all

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
```diff
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
...
+      let () =
+        match xes with
+        | [] -> ()
         | _ -> raise (Failure "Invalid symbolic")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
+      let cmd1 = (metadata, None, LLogic (LCmd.FreshSVar x_v)) in
+      let x_v = PVar x_v in
```
PetarMax, line 2144: So we're definitely not checking if `x_v` is actually a `PVar` name, rather than, say `#imapvar`?

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
```diff
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
...
+        | [] -> ()
         | _ -> raise (Failure "Invalid symb_number")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
-        (metadata, None, LLogic (LCmd.AssumeType (x, Type.NumberType)))
+      let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
+      let x_v = PVar x_v in
```
PetarMax, line 2169: Same as above...

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
```diff
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
...
+        | [] -> ()
+        | _ -> raise (Failure "Invalid symb_string")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
-        (metadata, None, LLogic (LCmd.AssumeType (x, Type.StringType)))
+      let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
+      let x_v = PVar x_v in
```
PetarMax, line 2184: Same as above-above...

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
```diff
@@ -2134,73 +2134,73 @@ let rec translate_expr tr_ctx e :
...
+        | [] -> ()
         | _ -> raise (Failure "Invalid symb_bool")
       in
       let x_v = fresh_var () ^ "_v" in
-      let cmd1 = (metadata, None, LBasic (BCmd.Assignment (x_v, LVar x))) in
-      let cmd2 = (metadata, None, LLogic (LCmd.SpecVar [ x ])) in
-      let cmd3 =
-        (metadata, None, LLogic (LCmd.AssumeType (x, Type.BooleanType)))
+      let cmd1 = (metadata, None, LLogic (FreshSVar x_v)) in
+      let x_v = PVar x_v in
```
PetarMax, line 2199: Guess what?

GillianCore/GIL_Syntax/LCmd.ml
```diff
@@ -8,8 +8,8 @@ type t = TypeDef__.lcmd =
   | Macro of string * Expr.t list  (** Macro            *)
   | Assert of Formula.t  (** Assert           *)
   | Assume of Formula.t  (** Assume           *)
-  | AssumeType of string * Type.t  (** Assume Type      *)
-  | SpecVar of string list  (** Spec Var         *)
+  | AssumeType of Expr.t * Type.t  (** Assume Type      *)
+  | FreshSVar of string  (** x := Symbol() *)
```
PetarMax, line 12: ...technically, `fresh_svar()` instead of `Symbpl()`...

PetarMax (changes requested):
Minor

Gillian-JS/lib/Compiler/JS2JSIL_Compiler.ml
giltho: So it can't be. In that case, that corresponds to a JS function call, so it has to be a JS program variable

PetarMax (dismissed):

PR #115, opened 2022-09-06
Debug unify: Add button to jump to subst's origin

NatKarmios:
Add a button to the debugger's unification interface, to jump to the assertion where a substitution is learned.
image

### DIFF
--- a/GillianCore/debugging/debugger/debugger.ml
+++ b/GillianCore/debugging/debugger/debugger.ml
@@ -104,11 +104,17 @@ struct
     type unify_kind = Unifier.unify_kind [@@deriving yojson]
     type unify_result = Success | Failure [@@deriving yojson]
 
+    type substitution = {
+      assert_id : rid; [@key "assertId"]
+      subst : string * string;
+    }
+    [@@deriving yojson]
+
     type assertion_data = {
       id : rid;
       fold : (rid * unify_result) option;
       assertion : string;
-      substitutions : (string * string) list;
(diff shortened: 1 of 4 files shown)


giltho:
Love it

GillianCore/debugging/debugger/debugger.ml
```diff
@@ -167,7 +173,12 @@ struct
           let asrt, _ = asrt_report.step in
           Fmt.str "%a" Asrt.pp asrt
         in
-        let substitutions = asrt_report.subst |> Subst.to_list_pp in
+        let substitutions =
+          asrt_report.subst |> Subst.to_list_pp
+          |> List.map (fun subst ->
+                 List.find_opt (fun prev -> prev.subst = subst) prev_substs
```
giltho, line 179: Let's not use polymorphic equal
`[%eq: (string * string)] prev.subst subst` should work

giltho (approved):
try that tiny change, if that works, push and we'll merge

giltho:
We need to think about how to test the debugger

philippagardner:
AWS code, found bug, is great.

We should think about what to do with the students: perhaps give them X algorithms, their job is to verify them in the tool and complete a form with questions….I’m busking this a bit.

There’s probably other ways of thinking about this….What happens inside AWS about testing the debugger…?

P




On 6 Sep 2022, at 02:11, Sacha Ayoun ***@***.******@***.***>> wrote:



We need to think about how to test the debugger

PR #112, opened 2022-08-18
Update Dockerfile

kraimalani:
upgrade esy to 0.6.12

PR #116, opened 2022-09-09
Update Dockerfile, add CI job to test Docker

NatKarmios:
Builds on #112


giltho:
I'll merge that, but it is horribly slow, I hadn't thought about that. We might only build the docker on release for example?

PR #117, opened 2022-09-21
Support "hidden" annotation

NatKarmios:
Allow commands to be annotated as hidden, causing them to not be considered when lifting to the target language.

### DIFF
--- a/GillianCore/GIL_Syntax/Annot.ml
+++ b/GillianCore/GIL_Syntax/Annot.ml
@@ -2,14 +2,10 @@
 type t = {
   origin_loc : Location.t option;  (** Better not to know what this is for *)
   origin_id : int option;  (** Origin Id, that should be abstracted away *)
-  loop_info : string list;
+  loop_info : string list; [@default []]
+  lift_hidden : bool; [@default false]  (** Hidden when lifting *)
 }
-[@@deriving yojson]
-
-(**/**)
-
-let make ?origin_loc ?origin_id ?(loop_info = []) () =
-  { origin_loc; origin_id; loop_info }
+[@@deriving yojson, make]
 
 let get_loop_info (annot : t) = annot.loop_info
 
@@ -18,3 +14,5 @@ let set_loop_info (annot : t) (loop_info : string list) =
 
 let get_origin_loc annot = annot.origin_loc
 let get_origin_id annot = annot.origin_id
+let hide (annot : t) = { annot with lift_hidden = true }
+let is_hidden (annot : t) = annot.lift_hidden
--- a/GillianCore/debugging/debugger/debugger.ml
+++ b/GillianCore/debugging/debugger/debugger.ml
@@ -297,6 +297,8 @@ struct
     }
     [@@deriving yojson]
 
+    type stop_at = StartOfPath | EndOfPath | BeforeNothing
+
     type 'case t =
       | Nothing
       | Cmd of cmd_data * 'case t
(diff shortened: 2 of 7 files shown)


GillianCore/GIL_Syntax/Annot.ml
```diff
@@ -2,14 +2,10 @@
...
+  loop_info : string list; [@default []]
+  lift_hidden : bool; [@default false]  (** Hidden when lifting *)
 }
-[@@deriving yojson]
-
-(**/**)
-
-let make ?origin_loc ?origin_id ?(loop_info = []) () =
-  { origin_loc; origin_id; loop_info }
+[@@deriving yojson, make]
```
giltho, line 8: nice

GillianCore/debugging/debugger/debugger.ml
```diff
@@ -535,82 +549,83 @@ struct
...
+              ("lifted_map", to_yojson branch_case_to_yojson lifted_map);
+              ("new_submap", to_yojson branch_case_to_yojson new_submap);
+            ])
+          ("ExecMap.insert_lifted: " ^ msg)
+      in
       let lift_data cmd_data =
         let new_display =
           Lifter.get_origin_node_str tl_ast cmd_data.origin_id
         in
-        let new_displa =
```
giltho, line 544: lmao

GillianCore/utils/syntaxes.ml
```diff
@@ -6,4 +6,5 @@ end
 module Option = struct
   let ( let+ ) f o = Option.map o f
   let ( let* ) o f = Option.bind o f
+  let ( let- ) f o = Option.iter o f
```
giltho, line 9: Yep, you figured out that wouldn't work I guess: iter doesn't compose like map and bind do.
bind is `'a t -> ('a -> 'b t) -> 'b t`
map is `'a t -> ('a -> 'b) -> 'b t`
Both take and return a `_ t` so you can compose those as you want.
iter is `'a t -> ('a -> ()) -> ()`, that doesn't compose, so you can't use monadic let binding.

Anyway, let's remove it :)

GillianCore/utils/option_utils.ml
```diff
@@ -11,3 +11,8 @@ let somes_and_eq ox y =
 let or_else f = function
   | Some x -> x
   | None -> f ()
+
+let coalesce a b =
+  match (a, b) with
+  | Some x, _ | _, Some x -> Some x
+  | _ -> None
```
giltho, line 18:
```suggestion
let coalesce a b =
  match a with
  | Some _ -> a
  | _ -> b
```
equivalent solution, but lighter.
First, matching on b is not necessary, so that's an operation you can avoid
Second, in the case when you return a, you don't have to rebuild a `Some x` as the `a` allocation may simply be reused

giltho: acknowledging that was my own code 😆

giltho (changes requested):

GillianCore/utils/syntaxes.ml
NatKarmios: Removed in b77c1f2afb712b4c68107ea9cf97b96714c4a3a2

GillianCore/utils/option_utils.ml
NatKarmios: Changed in 04099f5707f52ded44d805023b721f07f97650a2

PR #119, opened 2022-09-28
Support debugging multiple procs simultaneously

NatKarmios:
Not in the parallel sense - essentially allows context switching between multiple procs when debugging.

This is in preparation for debugging loops, which, in WISL, are abstracted away as function calls upon compilation.

### DIFF
--- a/GillianCore/debugging/debugger/debugger.ml
+++ b/GillianCore/debugging/debugger/debugger.ml
@@ -605,14 +605,9 @@ struct
         let new_display =
           Lifter.get_origin_node_str tl_ast cmd_data.origin_id
         in
-        if new_display = "No info!" then
-          DL.failwith
-            (fun () ->
-              let cmd_data =
-                cmd_data_to_yojson branch_case_to_yojson cmd_data
-              in
-              [ ("cmd_data", cmd_data) ])
-            "HORROR: tried to lift hidden command!";
+        let new_display =
+          if new_display = "No info!" then cmd_data.display else new_display
+        in
         { cmd_data with display = new_display }
       in
       let eq_opt ida idb =
--- a/GillianCore/debugging/debugger/debugger.ml
+++ b/GillianCore/debugging/debugger/debugger.ml
@@ -681,6 +681,7 @@ struct
     mutable lifted_exec_map : branch_case ExecMap.t option;
     mutable proc_name : string option;
     mutable unify_maps : (rid * UnifyMap.t) list;
+    report_state : L.report_state;
   }
 
   type debug_cfg = {
@@ -690,6 +691,7 @@ struct
     tl_ast : tl_ast option;
     tests : (string * Verification.t) list;
     main_proc_name : string;
+    report_state_base : L.report_state;
   }
 
   type debug_state = {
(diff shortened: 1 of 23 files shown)


giltho:
How can I experiment and assess if it's working?

giltho:
CI machine broke

NatKarmios:
> How can I experiment and assess if it's working?

@giltho, ok, I've reached a decent stopping point for the PR; the UI hasn't been updated, but we can see it working via the logged debug state (pictured)

image

GillianCore/logging/logging.mli
```diff
@@ -94,6 +94,14 @@ module LogQueryer : sig
...
 end
 
+type report_state
+
+val new_report_state : unit -> report_state
+val clone_report_state : report_state -> report_state
+val global_report_state : report_state
+val activate_report_state : report_state -> unit
+val with_report_state : (unit -> 'a) -> report_state -> 'a
+
```
giltho, line 104:
```ocaml
module Report_state = struct 
  type t
  val make: unit -> t
  val copy : t -> t
  val global : t
  val activate : t -> unit
  val with_report_state: (unit -> 'a) -> report_state -> 'a
end
```
or something similar? Whatever makes sense? Maybe do that in `ReportBuilder` directly, and then just alias the module?

GillianCore/engine/Abstraction/Verifier.ml
```diff
@@ -1045,8 +1047,9 @@ struct
     Printf.printf "%s\n" msg;
     L.normal (fun m -> m "%s" msg)
 
-  let verify_up_to_procs (prog : prog_t) :
+  let verify_up_to_procs ?(proc_name : string option) (prog : prog_t) :
       SAInterpreter.result_t SAInterpreter.cont_func =
+    proc_name |> ignore;
```
giltho, line 1052: ?

GillianCore/debugging/debugger/debugger.ml
```diff
@@ -650,27 +686,46 @@ struct
...
+    report_state_base : L.report_state;
   }
 
+  type debug_state = {
+    cfg : debug_cfg;
+    procs : (string, debug_proc_state) Hashtbl.t;
+    mutable cur_proc_name : string;
+  }
+
+  let get_proc_state ?proc_name ?(activate_report_state = true) dbg =
```
giltho, line 708: You added `?proc_name` to a lot of functions, but it seems that you're never using the parameter.
Regarding optional parameters, I prefer to add them as needed. They introduce new control flows, which is always nice to not do if possible

giltho (commented):
Small things

GillianCore/logging/logging.mli
NatKarmios: Resolved in 29a513adb40f0d75f819d837419964f12f3c370c

GillianCore/engine/Abstraction/Verifier.ml
NatKarmios: Resolved in c0f957eaf54d88a6a982d9dd569a5cc9a30f4d56

GillianCore/debugging/debugger/debugger.ml
NatKarmios: Resolved in c0f957eaf54d88a6a982d9dd569a5cc9a30f4d56

giltho:
Oopsie

PR #121, opened 2022-09-30
JSON-ify GIL AST

NatKarmios:
Also, dump all compiled procs to JSON and log when using the debugger.

### DIFF
--- a/GillianCore/GIL_Syntax/Cmd.ml
+++ b/GillianCore/GIL_Syntax/Cmd.ml
@@ -7,7 +7,7 @@
 
 module SS = Containers.SS
 
-type logic_bindings_t = string * (string * Expr.t) list
+type logic_bindings_t = string * (string * Expr.t) list [@@deriving yojson]
 
 type 'label t = 'label TypeDef__.cmd =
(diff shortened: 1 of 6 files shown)


giltho:
Ok, so one issue with all this: we're dependent on ppx_deriving_yojson's internal choices for serialization and deserialization
I'd like to be able to serialize a GIL AST in Rust and deserialize it in OCaml. How do I make sure things coincide?

NatKarmios:
> Ok, so one issue with all this: we're dependent on ppx_deriving_yojson's internal choices for serialization and deserialization I'd like to be able to serialize a GIL AST in Rust and deserialize it in OCaml. How do I make sure things coincide?

Should this be considered alongside the wider considerations concerning serialisation? If I recall correctly, we quoted JSON (de)serialisation as a bottleneck. Perhaps we should know what we're doing in that regard before worrying about platform consistency?

giltho:
Could we get measures on that? Time spent serialising and deserialising json things in db mode?
That being said, the answer is "possibly"
Can you take a look at if atdgen applies here?

Issue #131, opened 2022-10-20
Fix Gillian's Infra


giltho:
Gillian's infra is all over the place

First, BiState is actually not using a `PState`, it's just using a `SState`. So actually, the diagram is wrong.

Second, take the `State.init` function for example. It's actually only used for `SState` and `CState` before evaluating a program. It's not actually part of "every state".

We need to isolate and differentiate "Memory capabilities" (Action vs Consumer/Producers/Fixes)

It should be possible to incrementally implement a memory. A bit like traits, and you can pass a Memory to the functor that builds bi-abduction iff you have every capability, but you only need the action capability to run wpst.

We shouldn't ever have functions that say "not implemented for this kind of state". It's a sign that the architecture is wrong and misunderstood.

giltho:
`SState` and `CState` have the same "flavour", e.g. "executable states", the need an init function, which is not needed by `PState` and `Bistate`

PR #132, opened 2022-10-20
Add a notion of global environment to Gillian

giltho:
This PR adds a notion of _initialisation data_ to Gillian
When initialised, a memory receives an `init_data` parameter of its desired type, and is abled to initialise itself with context.

For example, in C, this avoids having to create a global environment dynamically. Instead, it is created statically when the file is compiled and then carried around throughout execution.

This looks big, because it filters throughout the entire engine: every time we create a state, we have to give it `init_data`, and every component of the engine has to agree on its type, which creates annoying but necessary type equations.

On the way, I tried to fix an inconsistency or two about the infrastructures, and took a few notes - see #131.

Before this gets merged, I would like to:
- [x] Clean the LAction types to get rid of anything related to Global env core predicates which don't exist anymore
- [x] Add a syntax for init_data in GIL. For now, I'll just add it as a section at the beginning of the file, it will either be whatever text that has to be parsed, or JSON. I'm thinking I'll go with JSON


giltho:
Ready for review and merge

Gillian-C/lib/MonadicSMemory.ml
```diff
@@ -466,44 +454,56 @@ module Mem = struct
...
+  else Mem.empty
+
+let init genv = { genv; mem = ref (just_functions genv) }
+let clear { genv; _ } = { genv; mem = ref (just_functions genv) }
+
+(* Getters *)
+
+(* Getters *)
+
+(* Getters *)
```
NatKarmios, line 504: ?

giltho: What the fuck hahahaha

NatKarmios (approved):
I don't see anything immediately horrifying

giltho:
Thank you, removing the weirdness you flagged and merging

Issue #142, opened 2022-11-02
Produce (Sphinx?) documentation


NatKarmios:
We want the website and docs to be one site.
Probably best to keep it all in this repo, and auto-push to the org .io page on merges to master.
Since the API's going to change so frequently, it's not really worth versioning it at the moment.

[Sphinx docs](https://www.sphinx-doc.org/en/master/)
[Sphinx ocaml extension](https://ocaml.org/p/sphinxcontrib-ocaml/0.2.0)

NatKarmios:
Apparently, the [drom](https://github.com/OCamlPro/drom) tool generates projects with documentation set up in a sensible way - it might be worth matching our docs setup with theirs.

PR #186, opened 2022-11-28
Set up documentation

NatKarmios:
Related to #142


esy.json
```diff
@@ -60,7 +60,15 @@
     "install-githooks": "./githooks/install.ml",
     "clean-githooks": "./githooks/clean.ml",
     "clean": "dune clean",
-    "watch": "dune build @install --promote-install-files --watch"
+    "watch": "dune build @install --promote-install-files --watch",
+    "odoc": "sh -c \"esy x dune build @doc && ./scripts/sync_odoc.sh > /dev/null\"",
+    "odoc:watch:build": "dune build @doc --watch",
+    "odoc:watch:sync": "./scripts/watch_odoc.sh > /dev/null",
+    "odoc:watch": "concurrently -p none \"esy odoc:watch:build\" \"esy odoc:watch:sync\"",
+    "sphinx": "sh -c \"sphinx-build sphinx _docs/sphinx/ \"",
```
giltho, line 68: Why is the `sh-c` needed here if you're not using `&&`?

sphinx/_extensions/gillian_custom_roles.py
```diff
@@ -0,0 +1,195 @@
+# pylint: disable=dangerous-default-value
```
giltho, line 1: That's a cool file :)

giltho (approved):

scripts/watch_odoc.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+inotifywait -qmr -e modify,create,delete ./_build/default/_doc/_html | xargs -n1 -I{} "./scripts/sync_odoc.sh"
```
giltho, line 2: Can we add a platform detection if/else to choose if we use fswatch or inotifywait?
Because I don't have the latter on Mac

giltho, line 2: Can we add a platform detection if/else to choose if we use fswatch or inotifywait?
Because I don't have the latter on Mac

giltho (changes requested):

scripts/watch_odoc.sh
NatKarmios: Resolved in f6619e5a4b519e150c594f4a5f96340b068b397f

esy.json
NatKarmios: It isn't :P 
3b05c801d9b171b5d36eb847d3e09754f7be3816

PR #189, opened 2022-11-28
Clean up debugger code

### LINKED ISSUE #139: Evaluate `Debugger_utils`
From #134 

The contents of `Debugger_utils` are very much an artefact of past development - I haven't touched them since starting work on Gillian, aside from giving them their own module.

We ought to take another look and see just how necessary all of it is, since the shape of the debugger has changed a lot since their induction.

Supposedly this might not be doable until JS is fixed (#113).

### LINKED ISSUE #141: Make `ExecMap.unifys` more sensible
From #134 

```ocaml
type unifys = (rid * Unifier.unify_kind * UnifyMap.unify_result) list
```

should probably be

```ocaml
type unify = {
  id : rid;
  kind : Unifier.unify_kind;
  result : UnifyMap.unify_result;
}
```

### LINKED ISSUE #143: Add more comments to `ExecMap` and `UnifyMap`
From #134 

`ExecMap` and `UnifyMap` just need a whole bunch of comments. More people than just me ought to know how they work :))

### LINKED ISSUE #158: Overhaul code structure
We could be using some more succinct code structures.

The two main techniques to apply here would be:
- "[The `_intf` trick](https://www.craigfe.io/posts/the-intf-trick)" - this removes a lot of duplication between `.ml` and `.mli` files
- Use destructive substitution for constraint-heavy functors - see [here](https://dev.realworldocaml.org/functors.html) for more info

NatKarmios:
Resolves #139
Resolves #141
Resolves #143
Resolves #158


scripts/watch_odoc.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
 if [ -x "$(command -v inotifywait)" ]; then
+  echo "Watching with inotifywait"
   inotifywait -qmr -e modify,create,delete ./_build/default/_doc/_html | xargs -n1 -I{} "./scripts/sync_odoc.sh"
 elif [ -x "$(command -v fswatch)" ]; then
   fswatch -o ./_build/default/_doc/_html | xargs -n1 -I{} "./scripts/sync_odoc.sh"
-else 
+else
+  echo "Watching with fswatch"
```
giltho, line 9: woopsie

giltho (changes requested):

scripts/watch_odoc.sh
NatKarmios: wops
fed444d75b0519c548ea6f1961a1949466c991e5

GillianCore/debugging/utils/exec_map.ml
```diff
@@ -1,38 +1,62 @@
 module L = Logging
```
giltho, line 1: How about some module-level documentation about what an exec map is, like what it's for, what it looks like.
The annotations on the type lack context

GillianCore/debugging/utils/unify_map.ml
```diff
@@ -2,43 +2,51 @@ module L = Logging
 module DL = Debugger_log
```
giltho, line 2: Same as exec_map, you can even reference exec_map

giltho:
But at the same time having sphinx makes it a LOT more readable so 🤷

giltho (approved):

PR #214, opened 2023-02-17
check sat with only relevant information in monadic mode

giltho:
Signed-off-by: Sacha Ayoun <sachaayoun@gmail.com>


giltho:
I had a revelation that this change would increase perfs, and it slightly does

NatKarmios (approved):

NatKarmios:
For reference, this shouldn't be merged before #229

NatKarmios (approved):

Issue #199, opened 2022-12-13
Cover all public-facing modules in documentation


NatKarmios:
- [ ] `Gil_syntax`
- [ ] `Gil_parsing`
- [ ] `CommandLine`
- [ ] `Symbolic`
- [ ] `Concrete`
- [ ] `General`
- [ ] `Bulk`
- [ ] `Monadic`
- [ ] `Debugger`
- [ ] `Logic`
- [ ] `Abstraction`
- [ ] `Logging`
- [ ] `IncrementalAnalysis`
- [ ] `Utils`

NatKarmios:
This would be nice, but nobody's going to have time for this before next-gen Gillian.

PR #217, opened 2023-02-22
Refactor and document `Command_line`

### LINKED ISSUE #126: Split `CommandLine.ml` into multiple files

NatKarmios:
Contributes to #199 
Resolves #126


GillianCore/parserAndCompiler/ParserAndCompiler.mli
```diff
@@ -8,6 +10,12 @@ type ('annot, 'tl_ast, 'init_data) compiled_progs = {
   init_data : 'init_data;
 }
 
+val get_progs_or_fail :
+  (Format.formatter -> 'e -> unit) ->
+  (('a, 'b, 'c) compiled_progs, 'e) result ->
+  ('a, 'b, 'c) compiled_progs
```
giltho, line 16: Some documentation on this function
Also, the first argument is quite surprising, I would make it labeled, `~pp_err`
My rule of thumb for making arguments labeled is: do I already know what the argument is given the function name and the argument type?

```ocaml
(* Doesn't need labels *)
val use_query: database -> query -> query_result

(* Wait, what's this weird first argument? Better make it labeled *)
val use_query: (string -> string) -> database -> query -> query_result

(* Arguments have the same type, how do I know which one is which? *)
val copy_content: channel -> channel -> unit

(* This is better *)
val copy_content : in:channel -> out:channel -> unit
```

giltho (changes requested):
A tiny change, otherwise ready to merge
This is great btw

GillianCore/parserAndCompiler/ParserAndCompiler.mli
NatKarmios: Very good point, well made :)
So well made, in fact, that I'm going to add that to the style guide!

Made the change in 3d2cc07.

giltho (approved):

Issue #177, opened 2022-11-17
Think about testing for debugger and/or webview


NatKarmios:
Integration tests (e.g. cypress) may be possible, though we'd have be able to render the webview independently of VSCode functionality (or make all that functionality mockable).

Should the debugger be tested internally (i.e. within OCaml) or externally (i.e. with an external tool running the debugger)? Does this external tool exist?

PR #221, opened 2023-03-07
Enhance map views in debugger interface

### LINKED ISSUE #170: Enhance unify map
- [x] Show folds as nested maps instead of stepping in and out
      #221
- [x] When an assertion fails, show that assertion as the failure, rather than a "Failure" node beneath it
      #221
- [x] ~~Update the state to show what's been consumed as unification progresses~~
      Moved to #220

### LINKED ISSUE #176: Add show/hide button for ExecMap/UnifyMap expansions

NatKarmios:
Resolves #176 
Resolves #170 

- Allow nested exec/unify maps to be collapsed
- Render unification folds as nested maps
- Absorb Success/Failure unify nodes into the previous assertion

### DIFF
--- a/debugger-vscode-extension/package.json
+++ b/debugger-vscode-extension/package.json
@@ -28,7 +28,7 @@
   },
   "scripts": {
     "extension:build": "webpack",
-    "extension:watch": "tsc -watch -p ./",
+    "extension:watch": "tsc --watch --preserveWatchOutput -p ./",
     "webview:build": "TAILWIND_MODE='build' snowpack build",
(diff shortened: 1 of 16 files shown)


giltho (approved):
We really need a way of testing this, or simplifying the review of this kind of code

NatKarmios:
> We really need a way of testing this, or simplifying the review of this kind of code

This is #177. Can make it a priority if need be :)

giltho:
Priorities will be assessed when GJS is fixed

PR #231, opened 2023-03-23
add a vanish function to Delayed, use carefully

giltho:


giltho:
@NatKarmios  Imma self review that one if that's ok with you hahahaha

PR #229, opened 2023-03-15
Fix JS

NatKarmios:
Resolves #113

PR #240, opened 2023-04-04
[INGORE] Figuring out Fix-JS CI

NatKarmios:
This is just to play with the CI and figure out where #229 is going wrong


NatKarmios:
Don't need this anymore

PR #246, opened 2023-05-02
`('a list, 'b list) result => ('a, 'b) result list` as signature of execute_action

giltho:
For now, it comes with a `Legacy_s_memory.Modernize` wrapper around the old version of memory actions, so no behaviour change, but that's the biggest part of the step


NatKarmios:
Debugger seems to be unscathed :)

NatKarmios (approved):

PR #250, opened 2023-05-26
Gavin/ptr chunk v2

ggwg:
# Pattern match based on the type of the chunk

**Test Program 1**
```
struct IntPtrWrapper {
    int *age;
};

int *getPtrFromStruct(struct IntPtrWrapper *a) {
    int *age = a->age;
    return age;
}
```

**Test Program 2**
```
typedef struct node_s {
    int val;
    struct node_s *next;
    struct node_s *prev;
} Node;

typedef struct list_s {
    Node *head;
    Node *tail;
} List;

int test(List *list) {
    Node *node = list->head;
    // Node *next = node->next;
    // return next->val;
    return 0;
}
```

---

**Example Output of load operation with matched pointer type**
```
--test: 2--
TIME: 0.059568
CMD: gvar__1 := "i__loadv"("ptr", gvar__0)
LOC: unknown loc
PROCS: [test]
LOOPS: [] ++ []
BRANCHING: 0
```
- Note: the loaded type is a "ptr", which will then be changed to Chunk.Mptr type.


ggwg:
**TODO:**
- Apply the right fix for Mptr type
```
| Mfloat32 ->
            (Expr.EList [ Expr.string single_type; new_var_e ], Type.NumberType)
        | Mfloat64 ->
            (Expr.EList [ Expr.string float_type; new_var_e ], Type.NumberType)
        | Mint64 ->
            (Expr.EList [ Expr.string long_type; new_var_e ], Type.IntType)
        (* TODO: Change this *)
        | Mptr ->
            if Compcert.Archi.ptr64 then
              (Expr.EList [ Expr.string long_type; new_var_e ], Type.IntType)
            else (Expr.EList [ Expr.string int_type; new_var_e ], Type.IntType)
        | _ -> (Expr.EList [ Expr.string int_type; new_var_e ], Type.IntType)
```

**Things to double check** 
- Correctness of pattern matching:
```
(* Lookup type of variable using Clight program to determine whether it is a pointer *)
      let chunk =
        match compcert_chunk with
        (* Check whether chunk is a pointer - Mint64 or Mint32 can be pointer types *)
        | Mint64 | Mint32 -> (
            let typ =
              match expp with
              (* Get the id of the variable from CSharpMinor *)
              | Ebinop (_, Evar id, _) ->
                  let open Clight in
                  let rec find_var var_list =
                    match var_list with
                    | (var_id, var_typ) :: vs ->
                        if var_id == id then var_typ else find_var vs
                    | _ ->
                        failwith
                          (Printf.sprintf
                             "Variable with ident %d was not found inside \
                              function %s of Clight"
                             (num_to_integer id) fname)
                  in
                  let clight_fun =
                    Gil_logic_gen.get_clight_fun clight_prog fid
                  in
                  find_var
                    (clight_fun.fn_params @ clight_fun.fn_vars
                   @ clight_fun.fn_temps)
              | _ -> failwith (Printf.sprintf "Unexpected type")
            in
            match typ with
            | Tpointer _ -> Chunk.Mptr
            | _ -> Chunk.of_compcert_ast_chunk compcert_chunk)
        | _ -> Chunk.of_compcert_ast_chunk compcert_chunk
```

Gillian-C/lib/Chunk.ml
```diff
@@ -7,10 +7,31 @@ type t = Compcert.AST.memory_chunk =
...
+  | Mint8unsigned -> Mint8unsigned
+  | Mint16signed -> Mint16signed
+  | Mint16unsigned -> Mint16unsigned
+  | Mint32 -> Mint32
+  | Mint64 -> Mint64
+  | Mfloat32 -> Mfloat32
+  | Mfloat64 -> Mfloat64
+  | Many32 | Many64 -> Mptr (* Should we fail with error here instead? *)
+
+let to_compcert_ast_chunk : t -> Compcert.AST.memory_chunk = function
```
giltho, line 24: let's call those `to_compcert` and `of_compcert`

Gillian-C/lib/Chunk.ml
```diff
@@ -33,21 +53,55 @@ let to_string = function
...
 
 let pp fmt chunk = Fmt.pf fmt "%s" (to_string chunk)
-let type_of = Compcert.AST.type_of_chunk
+
+let type_of = function
+  | Mint64 -> Compcert.AST.Tlong
+  | Mfloat32 -> Compcert.AST.Tsingle
+  | Mfloat64 -> Compcert.AST.Tfloat
+  | Mptr ->
+      if Compcert.Archi.ptr64 then Compcert.AST.Tany64 else Compcert.AST.Tany32
```
giltho, line 65: I think the right thing to do here is to return `Tlong` when `ptr64` is true and `Tint` otherwise. That's previous behaviour, expected by the memory model

Gillian-C/lib/Chunk.ml
```diff
@@ -33,21 +53,55 @@ let to_string = function
...
+    | Mint16signed -> Memdata.size_chunk Mint16signed
+    | Mint16unsigned -> Memdata.size_chunk Mint16unsigned
+    | Mint32 -> Memdata.size_chunk Mint32
+    | Mint64 -> Memdata.size_chunk Mint64
+    | Mfloat32 -> Memdata.size_chunk Mfloat32
+    | Mfloat64 -> Memdata.size_chunk Mfloat64
+    | Mptr ->
+        if Compcert.Archi.ptr64 then Zpos (Coq_xO (Coq_xO (Coq_xO Coq_xH)))
+        else Zpos (Coq_xO (Coq_xO Coq_xH))
+  in
```
giltho, line 84: `to_compcert chunk |> Memdata.size_chunk |> Camlcoq.Z.to_int` would be more succint and less error prone here :)

Gillian-C/lib/Chunk.ml
```diff
@@ -33,21 +53,55 @@ let to_string = function
...
+    | Mint8signed -> Memdata.align_chunk Mint8signed
+    | Mint8unsigned -> Memdata.align_chunk Mint8unsigned
+    | Mint16signed -> Memdata.align_chunk Mint16signed
+    | Mint16unsigned -> Memdata.align_chunk Mint16unsigned
+    | Mint32 -> Memdata.align_chunk Mint32
+    | Mint64 -> Memdata.align_chunk Mint64
+    | Mfloat32 -> Memdata.align_chunk Mfloat32
+    | Mfloat64 -> Memdata.align_chunk Mfloat64
+    | Mptr -> Zpos (Coq_xO (Coq_xO Coq_xH))
+  in
```
giltho, line 103: Same as size_chunk

Gillian-C/lib/Chunk.ml
```diff
@@ -33,21 +53,55 @@ let to_string = function
...
+  Camlcoq.Z.to_int value
 
-let equal = Compcert.AST.chunk_eq
-let ptr = Compcert.AST.coq_Mptr
+let equal chunk1 chunk2 = chunk1 = chunk2
```
giltho, line 106: Please use `[@@deriving eq]` on `type t` instead.
Polymorphic equality is much slower than using an actual function that explains how to test for equality in ocaml

Gillian-C/lib/cParserAndCompiler.ml
```diff
@@ -282,12 +282,12 @@ let parse_and_compile_file path exec_mode =
   let last_clight = get_or_print_and_die (SimplLocals.transf_program clight) in
   let csm = get_or_print_and_die (Cshmgen.transl_program last_clight) in
   let () =
-    if !Config.burn_csm then
-      let pathcsm = Filename.chop_extension path ^ ".csm" in
-      let oc = open_out pathcsm in
-      let fmt = Format.formatter_of_out_channel oc in
-      let () = Format.fprintf fmt "%a" PrintCsharpminor.print_program csm in
-      close_out oc
+    (* if !Config.burn_csm then *)
```
giltho, line 285: Please revert that :)
Also, for next time, writing the csm down is activated by using `--burn-csm` on the CLI

Gillian-C/lib/gilgen.ml
```diff
@@ -130,18 +130,110 @@ let trans_binop_expr ~fname binop te1 te2 =
...
+                    match var_list with
+                    | (var_id, var_typ) :: vs ->
+                        if var_id == id then var_typ else find_var vs
+                    | _ ->
+                        failwith
+                          (Printf.sprintf
+                             "Variable with ident %d was not found inside \
+                              function %s of Clight"
+                             (num_to_integer id) fname)
+                  in
```
giltho, line 221: `==` tests for physical equality in OCaml. But two equal ids are not guaranteed to be physically equal (as in one could be a copy of the other). Use `Camlcoq.P.eq` instead
Also,
```ocaml
let all_vars = clight_fun.fn_params @ clight_fun.fn_vars @ clight_fun.fn_temps in
let typ =
  match List.find_map (fun (var_id, var_typ) -> if Camlcoq.P.eq id var_id then Some var_typ else None) all_vars with
  | Some t -> t
  | None -> Fmt.failwith "Variable with ident %d ...." ...
in 
...
```

giltho (changes requested):
A few things, I think it's good otherwise, but it requires testing it :)

ggwg:
Thank you so much for reviewing on the bank holiday!

Gillian-C/lib/cParserAndCompiler.ml
ggwg: Oh, good catch!

Gillian-C/lib/gilgen.ml
ggwg: That's very elegant

Gillian-C/lib/Chunk.ml
```diff
@@ -7,9 +7,30 @@ type t = Compcert.AST.memory_chunk =
...
+let of_compcert : Compcert.AST.memory_chunk -> t = function
+  | Mint8signed -> Mint8signed
+  | Mint8unsigned -> Mint8unsigned
+  | Mint16signed -> Mint16signed
+  | Mint16unsigned -> Mint16unsigned
+  | Mint32 -> Mint32
+  | Mint64 -> Mint64
+  | Mfloat32 -> Mfloat32
+  | Mfloat64 -> Mfloat64
+  | Many32 | Many64 -> Mptr (* Should we fail with error here instead? *)
```
giltho, line 22: Sorry more comments
Yes, that should fail here, but also, what is `of_compcert` actually doing?
To convert from compcert to your new kind, you need the context?

Gillian-C/lib/gil_logic_gen.ml
```diff
@@ -155,7 +154,7 @@ let assert_of_member cenv members id typ =
     let n_e = Expr.int_z n in
     let chunk =
       match Ctypes.access_mode ty with
-      | By_value chunk -> chunk
+      | By_value chunk -> Chunk.of_compcert chunk
```
giltho, line 157: You have the type here, if it's a pointer type you should be using the `Mptr` chunk

Gillian-C/lib/gilgen.ml
```diff
@@ -130,18 +130,112 @@ let trans_binop_expr ~fname binop te1 te2 =
...
+  | Sbuiltin _ -> Fmt.pf fmt "[stmt] Sbuiltin"
+  | Sseq (s1, s2) -> Fmt.pf fmt "[stmt] Sseq (%a, %a)" pp_stmt s1 pp_stmt s2
+  | Sifthenelse _ -> Fmt.pf fmt "[stmt] Sifthenelse "
+  | Sloop s -> Fmt.pf fmt "[stmt] Sloop (%a)" pp_stmt s
+  | Sblock s -> Fmt.pf fmt "[stmt] Sblock (%a)" pp_stmt s
+  | Sexit _ -> Fmt.pf fmt "[stmt] Sexit "
+  | Sswitch _ -> Fmt.pf fmt "[stmt] Sswitch "
+  | Sreturn _ -> Fmt.pf fmt "[stmt] Sreturn "
+  | Slabel (label, s) ->
+      Fmt.pf fmt "[stmt] Slabel (%d, %a)" (num_to_integer label) pp_stmt s
```
giltho, line 184: Do we still need the printers?
Also, your `num_to_integer` is bound to exist in the `Camlcoq` library

giltho (changes requested):

Gillian-C/lib/gilgen.ml
ggwg: Forgot to remove them after debugging

Gillian-C/lib/gilgen.ml
```diff
@@ -338,9 +460,10 @@ let rec trans_stmt ~fname ~context stmt =
   | Sgoto lab ->
       let gil_lab = trans_label lab in
       [ (annot_ctx context, None, Cmd.Goto gil_lab) ]
-  | Sstore (chunk, vaddr, v) ->
+  | Sstore (compcert_chunk, vaddr, v) ->
       let addr_eval_cmds, eaddr = trans_expr vaddr in
       let v_eval_cmds, ev = trans_expr v in
+      let chunk = Chunk.of_compcert compcert_chunk in
```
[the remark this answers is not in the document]
ggwg: Nice. I am adding it

Gillian-C/lib/gilgen.ml
giltho: No no, Camlcoq is extracted from the coq development of compcert
In particular here, [`label`](https://compcert.org/doc/html/compcert.cfrontend.Csharpminor.html#label) is defined as [`ident`](https://compcert.org/doc/html/compcert.common.AST.html#ident) which is defined as [`positive`](https://compcert.org/doc/html/compcert.common.AST.html#ident) (the coq development is clickable)

A quick look at `Camlcoq` shows that the `P` module, for `positive` values contains a `to_int` function (which is used is [`trans_label`](https://github.com/GillianPlatform/Gillian/blob/master/Gillian-C/lib/gilgen.ml#L191) for example.

So you can replace all calls of `num_to_integer` by `Camlcoq.P.to_int`

Gillian-C/lib/Chunk.ml
ggwg: `of_compcert` is needed when translating the compcert chunks from the CSM program into out internal Chunk representation, called during `ELoad` and `SStore` functions.

PR #251, opened 2023-05-30
Flip arguments to `bind` when unfolding

### LINKED ISSUE #228: Reverse order of `bind` when unfolding
Currently, an unfold with `[bind (#a := #b)]` will set `#b` to `#a`, which defies intuition.

NatKarmios:
Resolves #228


giltho (approved):
It's passing the tests so let's merge :)

PR #273, opened 2023-11-30
Rust catch-up

### LINKED ISSUE #225: Consumers and producers shouldn't be actions
I want to rewrite the memory model interface to add a notion of consumer and producer directly as functions.
Signatures would be slightly different. For example producers shouldn't be allowed to fail, and do not return anything.

Of course, that would mean removing the notion of get/set/rem in favour of cons/prod, simplifying the interface

giltho:
Many things here:
1) Adds the magic wand operator the `Asrt.t`, and the `Package` operations, which creates a magic wand from the current state.
2) Implements aggregate unification plans (or master unification plans), revamping a big part of `UP.ml`
3) Adds implication to the formula language, which can be encoded as such to z3
4) Replace get/set/rem by consumers and consumers, fixes #225 
4) Adds support for list_repeat operator

Not everything should be merged possibly, let's trim some useless prints etc


giltho:
@NatKarmios This can now be reviewed and debugger can be tested independently of the CI passing if it's because of the z3 flakyness

NatKarmios (commented):
LGTM. Just checking debugger still works.

NatKarmios (approved):
Yup, debugger looks good

PR #276, opened 2024-02-01
Upgrade to OCaml 5

NatKarmios:
- [x] Resolve dependencies
- [x] Make compilation succeed
- [x] Make tests pass
- [x] Remake githook scripts (`shexp` has been replaced with ~`bos`~ `feather`)

Note that older versions of **esy** will break during dependency resolution. Currently confirmed to work on version `0.7.2`.


giltho (approved):
Surprisingly minor :o

PR #277, opened 2024-02-12
Rename unify to matching, convert state types to records

### LINKED ISSUE #3: Inline records for complicated types
There are several variant types that are extremely complicated and contain a lot of different parameters that are unnamed. Using [inline records](https://blog.janestreet.com/ocaml-4-03-everything-else/) would improve readability

### LINKED ISSUE #196: Rename Unification to Matching
This is not a high priority and will be annoying to do, but it ought to be done to get more in line with the theory.

NatKarmios:
- Resolves #196
- Resolves #3


giltho (approved):
Before merging I'd like to run bi-abduction and check that the results are those we had in the CSE paper, the rewriting you made is quite error-prone given state and af state are the same type, and that can cause spurious bugs e.g. when using record update syntax

NatKarmios:
> Before merging I'd like to run bi-abduction and check that the results are those we had in the CSE paper, the rewriting you made is quite error-prone given state and af state are the same type, and that can cause spurious bugs e.g. when using record update syntax

(we tested, and it's consistent with master)

PR #288, opened 2024-05-07
Allow empty postcondition+vanishing function calls

N1ark:
Allows the verification of vanishing function calls and empty (false) post-conditions.


giltho:
@NatKarmios Let's just change print_to_all to only print to stdout if not in debugger mode?

PR #304, opened 2024-07-17
Save init data when compiling + sorted outputs

N1ark:
- Ensure init data is also saved to the GIL file when using `compile`
- Sort outputs of the `act` command (allows diffing between executions)
- Fix Gillian-C not running with precompiled code if init data is not specified


NatKarmios (approved):
lgtm :) thanks

PR #309, opened 2024-08-08
add consume & produce commands

giltho:
Signed-off-by: Sacha Ayoun <sachaayoun@gmail.com>


N1ark (approved):
lgtm :)

would be nicer to define some shared function for consume + SepAssert but if it's a temporary thing it doesn't really matter

(edit: oops i forgot i dont have approval rights)

giltho:
> would be nicer to define some shared function for consume + SepAssert

I tried but didn’t quite work, the two functions return different things, I’d need to return a substitution etc… it could be done but it’s not a clean cut

Issue #159, opened 2022-11-04
Split the `Asrt` type


giltho:
The `Asrt.t` type weird, an assertion is actually a free monoid of "Simple assertions", and the representations are mixed.
I'd like to split that type into
```ocaml
type simple =
   | GA of ...
   | Pure of ...
   ... 


type t = simple list
```
It would avoid weird cases in the code where the assertions are split and then the `Star` pattern is ignored

It would also be a great time to rename `GA` to `CorePred`

PR #313, opened 2024-09-14
Use `Asrt.t` instead of fixes!

N1ark:
- Remove `c_fix_t` from `SMemory`, use `Asrt.t` instead
- Remove `fix_t` from `State`, use `Asrt.t` instead
- Simplify the signature of `get_fixes`: `err_t -> (Asrt.t list * Containers.SS.t) list` (ie. remove PFS, TypeEnv and State; the error should be enough to generate fixes)
- Remove `apply_fix`, `pp_fix_t`, `fix_t`
- Adapt Gillian instantiations to match

(ping @NatKarmios / @giltho!)

### DIFF
--- a/Gillian-C/lib/MonadicSMemory.ml
+++ b/Gillian-C/lib/MonadicSMemory.ml
@@ -903,7 +903,6 @@ let get_fixes err =
           let new_var_e1 = Expr.LVar new_var1 in
           let new_var2 = LVar.alloc () in
           let new_var_e2 = Expr.LVar new_var2 in
-          let set = SS.add new_var2 (SS.singleton new_var1) in
           let value = Expr.EList [ new_var_e1; new_var_e2 ] in
           let null_typ =
             if Compcert.Archi.ptr64 then Expr.string long_type
@@ -911,13 +910,12 @@ let get_fixes err =
           in
           let null_ptr = Expr.EList [ null_typ; Expr.int 0 ] in
           [
-            ( [
-                single ~loc ~ofs ~chunk ~sval:value ~perm;
-                Asrt.Types
-                  [ (new_var_e1, Type.ObjectType); (new_var_e2, Type.IntType) ];
-              ],
-              set );
-            ([ single ~loc ~ofs ~chunk ~sval:null_ptr ~perm ], SS.empty);
+            [
+              single ~loc ~ofs ~chunk ~sval:value ~perm;
+              Asrt.Types
+                [ (new_var_e1, Type.ObjectType); (new_var_e2, Type.IntType) ];
+            ];
+            [ single ~loc ~ofs ~chunk ~sval:null_ptr ~perm ];
           ]
       | _ ->
           let type_str, type_gil =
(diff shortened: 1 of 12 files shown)


giltho:
Why are SpecVar fixes required?

giltho:
But also thank you, I wanted to have this forever! This is really nice

N1ark:
> Why are SpecVar fixes required?

Well I'm not 100% sure they're needed; they just list the logical variables that need to be kept for the specification and can't be removed/renamed, I think? Didn't try to get rid of them

N1ark:
Hmm turns out SpecVars was apparently not needed? When running my biab tests I get the same results for WISL+C except that one postcondition for one C spec has one less logical variable, which is not used anyway:
```diff
  spec listPrepend (x, z)
    [[ (x == #x) *
       (z == #z) *
       (#x == {{ _$l_40, _lvar_7 }}) *
       (#z == {{ "long", 0i }}) *
       (0i i<=# _lvar_7) *
       (0i i<=# (l-len #x)) *
       (0i i<=# (l-len #z)) *
       types(#x : List, _lvar_7 : Int, _lvar_34 : Int, #z : List) *
       types(_lvar_34 : Int) *
       <mem_single>(_$l_40, (8i i+ _lvar_7), "ptr"; {{ _$l_39, _lvar_34 }}, "Freeable") ]]
    [[ (ret == #x) *
       (#x == {{ _$l_35, _lvar_7 }}) *
       (#z == {{ "long", 0i }}) *
-      (_lvar_33 == _$l_36) *
       (_lvar_8 == _$l_35) *
       (0i i<=# _lvar_7) *
       (0i i<=# (l-len #x)) *
       (0i i<=# (l-len #z)) *
       types(#x : List, _lvar_7 : Int, _lvar_8 : Obj, _lvar_34 : Int, 
+            #z : List) *
-            #z : List, _lvar_33 : Obj) *
       types(l-nth(#z, 1i) : Int) *
       <mem_single>(_$l_35, (8i i+ _lvar_7), "ptr"; {{ "long", l-nth(#z, 1i) }}, "Freeable") ]]
```

N1ark:
So basically I removed everything fix-related and now fixes are simple assertions :) The only thing is I throw an error when the fix encountered is a Wand/Pred since I'm not really sure these are ok fixes anyways (in particular Wand; does it make any sense?)

giltho:
It looks like the specvars are needed

This is master for bi-abduction on collections-c:
image

This is current version of this PR:
image

This is version of this PR with the SpecVar still there:
image

giltho:
I'm impressed at how you get EXACTLY the same results by using Asrt.t as fixes

giltho:
What I'm thinking is that SpecVars shouldn't be fixed, but they still need to be properly added:
any logical variable that makes it to the pre-condition should be added as a specvar in the state, at risk of disconnecting things otherwise.
So we don't need specific specvar fixes, it's just that when you're producing a given `fix: Asrt.t`, you should get `Asrt.lvars fix` and add those to the specvars, in both the current state and anti-frame

giltho:
What I'm running, btw, is :
```sh
opam exec -- dune exec -- gillian-c act \
  ../collections-c-for-gillian/gillian-compcert/biabduction/array/array_test_add.c \
  -I ../collections-c-for-gillian/gillian-compcert/utils \
  -I ../collections-c-for-gillian/libs/fixed-bi/include \
  -S ../collections-c-for-gillian/gillian-compcert/utils \
  -S ../collections-c-for-gillian/libs/fixed-bi \
  --ignore-undef --allocated-functions --ignore-multdef \
  --bi-unroll 3 -l disabled
  ```
  
  where [collections-c-for-gillian](https://github.com/GillianPlatform/collections-c-for-gillian) is cloned next to Gillian

N1ark:
Oh wow just 2 lines of code were enough to re-add the spec vars automatically! Cool now same results as before

N1ark:
Also fyi I have a branch to try to remove `Star` from `Asrt.t` and basically do as #159 suggests because it seemed like a related simple change, currently running into a few issues with the C instantiation and AWS but hopefully should be mergeable soon :)

giltho:
👀👀👀👀👀👀

giltho:
Merging when CI's done

PR #316, opened 2024-10-10
Remove Asrt.Star (fix #159)

### LINKED ISSUE #159: Split the `Asrt` type

N1ark:
Remove `Asrt.Star`; what used to be `Asrt.t` is now `Asrt.atom` (atomic assertions), and `Asrt.t` is now `atom list`, avoiding the need for recursive functions when traversing assertions, and making the distinction much clearer and simpler.
Rename `Asrt.GA` -> `Asrt.CorePred`


giltho:
I think if you fix this problem, it will also fix #301

GillianCore/GIL_Syntax/Asrt.ml
```diff
@@ -1,15 +1,19 @@
 (** {b GIL logic assertions}. *)
-type t = TypeDef__.assertion =
+type simple = TypeDef__.assertion_simple =
```
giltho, line 2: Can we call this type `atom`?

N1ark:
> I think if you fix this problem, it will also fix #301 This PR also re-orders the fields and it breaks something

Ah yes interesting you seem to have stumbled upon similar things... When I have more time (🫠) I'll look; what's odd is that the version that works *doesn't* sort the assertions, so I'm not certain how sorting then (Types>Pure>CorePred>Wand) causes the issue; will update 

And yes sure I'll rename it to `atom` it's nicer

GillianCore/GIL_Syntax/Asrt.ml
N1ark: Done in c6a7574

N1ark:
JS works but something is still odd for Gillian-C, which ends up verifying less branches than it used to, possibly due to ordering issues. Will leave as is for now and maybe come back later to it...

N1ark:
The tests pass which is a miracle but from diffing some of my local tests I get a few mismatches so not ready to merge yet fyi

N1ark:
@giltho @NatKarmios fyi this is ready for review :)

After some evaluation, I get:
- 4.07% perf improvement on Collections-C
- AWS passes
- [One specific proc](https://github.com/GillianPlatform/Gillian/blob/b9ff39f89280d97d3f02b0efb1234c2c707e9f75/Gillian-C/examples/verification/dll.c#L34-L62) is ~3 times slower, because more branches (77 rather than 13) are covered than before. I checked and they're ok, it's caused by some assertions being re-ordered and Gillian-C used to incorrectly die because of ALoc mismatches. I wonder if there's a way of avoiding this branch explosion though :/

Gillian-C/lib/MonadicSVal.ml
```diff
@@ -105,11 +105,16 @@ let of_gil_expr sval_e =
...
+    else
+      if%sat int_typ sval_e then DO.some (SVint (Expr.list_nth sval_e 1))
+      else
+        if%sat float_typ sval_e then DO.some (SVfloat (Expr.list_nth sval_e 1))
+        else
+          if%sat long_typ sval_e then DO.some (SVlong (Expr.list_nth sval_e 1))
+          else
+            if%sat single_typ sval_e then
+              DO.some (SVsingle (Expr.list_nth sval_e 1))
+            else DO.none ()
```
giltho, line 117: I think match%sat exists and does exactlyt that

NatKarmios (approved):
lgtm :)

Gillian-C/lib/MonadicSVal.ml
N1ark: Surprisingly, no there specifically is no `match%sat` -- is there a reason for it by the way? Otherwise I can just quickly implement it

`ppx_sat/expander/ppx_sat_expander.ml`:
```ocaml
let match_fun_fexpr_of_ext = function
  | Extension_name.Ent -> branch_entailment
  | Sat -> failwith "Should not get here, match should not be used with 'sat'"
```

giltho: I think my previous implementation of match%sat was unsound and I removed it?
You can make it syntactic sugar for a bunch of if%sat without implementing anything in delayed if you want

N1ark: Done in b4458b4 ! (Also wow ppx is sooo nice and easy to use??) I'll let you just review that and approve and it's ready to merge

NatKarmios: Lovely stuff :)

NatKarmios (approved):

N1ark:
broken chains

Issue #312, opened 2024-09-12
gillian-js is unable to produce failing model for modifed `Gillian-JS/Examples/Cosette/simple_example.js`


bjrjk:
I'm using the latest version of Gillian.
While the verification for original `simple_example.js` can be passed, after change the operator `+` to `*` according to https://gillianplatform.github.io/sphinx/js/symbolic-testing.html#assumptions-and-assertions , the verification cannot provide an counter example and crashes.
Changed `simple_example.js` is listed as follows:
```javascript
var n1 = symb_number(), n2 = symb_number();

Assume((0 <= n1) and (0 <= n2) and (not (n1 = n2)));

var res = n1 * n2; // changed + to *

Assert((n1 <= res) and (n2 <= res));

var x = symb();
Assume(not (typeOf x = Obj));

var tx = typeof(x);
```
Original `gillian-js` output:
```bash
~/SymbolicExecution/Gillian$ dune exec -- gillian-js wpst Gillian-JS/Examples/Cosette/simple_example.js
Compilation time: 0.010296s              
Total time (Compilation + Symbolic testing): 5.172468s
Success!
```

Failed `gillian-js` output is attached with an .txt due to its tremendous text length.
[failed_example.txt](https://github.com/user-attachments/files/16972817/failed_example.txt)

giltho:
Hi!
Well, the second assert shouldn't pass when using multiplication. Take n1 = 5 and n2 = 0.5, res is 2.5 and is less than n1. However, not quite sure why Gillian is failing to produce a simple model.
@NatKarmios could be your changes in the SMT connection, could you quickly investigate see what's happening and why we can't get a model?

bjrjk:
Thanks a lot!

klausnat:
Hello, yes, I am having very same problem... this is what I see instead of model

![example](https://github.com/user-attachments/assets/da5c4950-ff73-4d0a-aef5-14ecfc707064)

giltho:
I'll investigate at some point, sorry for the lack of responsiveness, my PhD defence is next week 😬

bjrjk:
No problem! As a PhD student, I understand your anxious feeling and that's totally understandable! The formal verification research is also very hard. Hope you will have a smooth defence and successfully graduate!

NatKarmios:
Many apologies for taking so long to look into this, I've been in deadline hell these past few months :-)
[...]

PR #318, opened 2024-10-25
fix issue #312: show counter example and failing model

klausnat:
After moving to [simple-smt-ocaml ](https://github.com/GaloisInc/simple-smt-ocaml) library, counter example and failing model wouldn't show up as described in [#issue 312](https://github.com/GillianPlatform/Gillian/issues/312).

This pull request fixes this problem. I added new function to the library (it takes configured solver as an argument).  
Please check attached files to examine "before fix" and "after fix" status

after_fix
before_fix

### DIFF
--- a/GillianCore/smt/smt.ml
+++ b/GillianCore/smt/smt.ml
@@ -157,9 +157,11 @@ let mk_datatype name type_params (variants : (module Variant.S) list) =
   let () = init_decls := recognizer_defs @ (decl :: !init_decls) in
   atom name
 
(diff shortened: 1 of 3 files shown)


NatKarmios:
Thank you very much for your PR @klausnat!
I now realise that the problem is that attempting to get the model will "re-declare" the built-in functions. Your PR fixes that, though it breaks cases where those functions are actually used.
The solution I've now found is to still provide those function declarations, but *after* the push to SMT state.
I'm including the fix in #322 🙂

PR #326, opened 2024-12-25
Remove `Formula`

N1ark:
###### (Not done, but opening as draft for now)

Partial Changelog: 
- Remove `Formula.t`, using `Expr.t` anywhere it was used
  - The only difference at this point was a slightly different AST and the fact an `Expr.t` is not always a boolan expression, but it's not actually that big of a deal.
  - The biggest downside I got so far is that it's slightly more verbose: `Eq (e1, e2)` becomes `BinOp (e1, Equals, e2)`, `Not f` becomes `UnOp (Not, e)`, etc.
- `BinOp`: rename `BAnd` → `And`, `BOr` → `Or`, `BImpl` → `Impl`, `SLessThan` → `StrLess`, `BSetMem` → `SetMem`, `BSetSub` → `SetSub`
- `UnOp`: rename `UNot` → `Not`, add `IsInt` from Formula
- Rename all symbolic infix operators (`#==`, `#<`, `#&&`...) to remove the hash, becoming `==`, `<`, etc. This deeply embeds the `Expr` operators which I found nicer and also means the operators actually have different precedence! Because all operators starting with `#` have [the same precedence](https://ocaml.org/manual/5.1/api/Ocaml_operators.html#top) which made using them annoying due to parenthesis.
- For GIL I didn't remove any of the existing syntax, but all `Formula` now compile as `Expr`, meaning that some operators have two equivalent symbols (eg. `=` and `==`, `<#` and `<`, `/\` and `&&`, `--e--` and `-e-`...)
- Remove the types `vt`, `st` and `action_ret` from the signature of the type `SMemory`, since their value was already set anyways (I think this doesn't change anything about the module type?) by using `type _ := _` instead.


GillianCore/engine/FOLogic/Reduction.ml
```diff
@@ -2412,638 +2636,84 @@ and substitute_for_list_length (pfs : PFS.t) (le : Expr.t) : Expr.t =
...
-               fe
-                 (Expr.Infix.( - )
-                    (Expr.list_length left_list)
-                    (Expr.list_length right_list))
-             with
-             | Expr.Lit (Int k) when not (Z.equal k Z.zero) -> true
-             | _ -> false ->
-          (* If we have two lists but can reduce the equality of their lengths to false,
-             then we know the lists cannot be equal*)
-          False
```
N1ark, line 2607: Removed this, which seems to slightly improve perf without causing any observable difference.

GillianCore/engine/FOLogic/Reduction.ml
```diff
@@ -2412,638 +2636,84 @@ and substitute_for_list_length (pfs : PFS.t) (le : Expr.t) : Expr.t =
...
-              | Some t1, Some t2 -> t1 <> t2
-              | _, _ -> false
-            then False
-            else
-              let ite a b : Formula.t = if a = b then True else False in
-              let default re1 re2 : Formula.t = Eq (re1, re2) in
-              match (re1, re2) with
-              (* DEDICATED RPFS REDUCTIONS *)
-              | NOp (LstCat, _), LVar y when rpfs && prefix_catch pfs re1 y ->
-                  Eq (UnOp (LstLen, re1), UnOp (LstLen, re2))
```
N1ark, line 2672: Removed this, as it incorrectly converted list equalities to list length equalities -- maybe because I removed the `rpfs` param? Either way it didn't seem to make that big of a difference 🤔

GillianCore/engine/FOLogic/Reduction.ml
```diff
@@ -1714,393 +1359,965 @@ and reduce_lexpr_loop
...
-                      PFS.exists
-                        (fun e ->
-                          Formula.equal e (Eq (flel, fler))
-                          || Formula.equal e (Eq (fler, flel)))
-                        pfs
-                    then Lit (Bool true)
-                    else if
-                      PFS.mem pfs (Not (Eq (flel, fler)))
-                      || PFS.mem pfs (Not (Eq (fler, flel)))
-                    then Lit (Bool false)
```
N1ark, line 1770: I had to remove these, as attempting to reduce `a = x` would always reduce to `true`, since that formula is already in the PFS. Maybe this reduction made sense when Expr/Formula were split? Note that a similar reduction for And/Or/Impl will still work -- this is the only case where it doesn't.

N1ark:
Some more work could be done but -- tests pass! Removing `Formula` and `reduce_formula` :)

Gillian-C/lib/MonadicSVal.ml
```diff
@@ -220,24 +231,22 @@ module SVArray = struct
...
         Logging.verbose (fun fmt ->
             fmt "Undefined pf: not as concrete: %a" Expr.pp size);
         let i = LVar.alloc () in
         let i_e = Expr.LVar i in
-        forall
-          [ (i, Some IntType) ]
-          zero #<= i_e #&& (i_e #< size)
-          #=> ((Expr.list_nth_e arr_exp i_e) #== (Lit Undefined))
+        forall [ (i, Some IntType) ] zero <= i_e
+        && i_e < size ==> (Expr.list_nth_e arr_exp i_e == Lit Undefined)
```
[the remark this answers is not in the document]
N1ark: Hi!

Thanks for checking it out :) Yes indeed I didn't notice, these expressions were wrong, thank you for pointing it out, just fixed it in 8111475. It was because the precedence of `#..` operators is[ higher than that of function application](https://ocaml.org/manual/5.1/api/Ocaml_operators.html) so when updating the syntax I missed it. (Un)luckily these bits of code were never run so I didn't notice :p

Yeah the change is very partially due by the ocamlformat change, though in general I think it makes more sense to define operators that have sensible precedence from the get go :)

PR #327, opened 2025-01-10
Remove `Formula`

N1ark:
Partial Changelog: 
- Remove `Formula.t`, using `Expr.t` anywhere it was used:
  The only difference at this point was a slightly different AST and the fact an `Expr.t` is not always a boolean expression, but it's not actually that big of a deal it seems.
  The biggest downside I got so far is that it's slightly more verbose: `Eq (e1, e2)` becomes `BinOp (e1, Equals, e2)`, `Not f` becomes `UnOp (Not, e)`, etc.
- Add patterns from `reduce_formula` into `reduce_lexpr`, cleaning up a bit, and remove `reduce_formula`
  **There are a few rules that I removed**, as they do not seem to stop the tests from passing and they caused errors after the merge -- I'll highlight these in the PR, to see if I'm correct in removing them.
- `BinOp`: rename `BAnd` → `And`, `BOr` → `Or`, `BImpl` → `Impl`, `SLessThan` → `StrLess`, `BSetMem` → `SetMem`, `BSetSub` → `SetSub`
- `UnOp`: rename `UNot` → `Not`, add `IsInt` from Formula
- Rename all symbolic infix operators (`#==`, `#<`, `#&&`...) to remove the hash, becoming `==`, `<`, etc. This deeply embeds the `Expr` operators which I found nicer and also means the operators actually have different precedence! Because all operators starting with `#` have [the same precedence](https://ocaml.org/manual/5.1/api/Ocaml_operators.html#top) which made using them annoying due to parenthesis.
- For GIL I didn't remove any of the existing syntax, but all `Formula` now compile as `Expr`, meaning that some operators have two equivalent symbols (eg. `=` and `==`, `<#` and `<`, `/\` and `&&`, `--e--` and `-e-`...)
  The only difference is that pure formulae must be enclosed in parenthesis, and all other assertions must not be in parenthesis, ie. `pred isTrue(+b: Boolean):  b = true;` must become `pred isTrue(+b: Boolean): (b = true);`. This is to allow distinguishing the assertion `*` operator from the expression's multiplication. (In reality this change only affected a handful of definitions).
- Remove the types `vt`, `st` and `action_ret` from the signature of the type `SMemory`, since their value was already set anyways (I think this doesn't change anything about the module type?) by using `type _ := _` instead.


GillianCore/engine/FOLogic/Reduction.ml
```diff
@@ -2485,566 +2614,11 @@ let resolve_expr_to_location (pfs : PFS.t) (gamma : Type_env.t) (e : Expr.t) :
...
-               fe
-                 (Expr.Infix.( - )
-                    (Expr.list_length left_list)
-                    (Expr.list_length right_list))
-             with
-             | Expr.Lit (Int k) when not (Z.equal k Z.zero) -> true
-             | _ -> false ->
-          (* If we have two lists but can reduce the equality of their lengths to false,
-             then we know the lists cannot be equal*)
-          False
```
N1ark, line 2608: Removed this because it has a significant impact of performance and didn't seem needed

GillianCore/engine/FOLogic/Reduction.ml
```diff
@@ -2485,566 +2614,11 @@ let resolve_expr_to_location (pfs : PFS.t) (gamma : Type_env.t) (e : Expr.t) :
...
-              | Some t1, Some t2 -> t1 <> t2
-              | _, _ -> false
-            then False
-            else
-              let ite a b : Formula.t = if a = b then True else False in
-              let default re1 re2 : Formula.t = Eq (re1, re2) in
-              match (re1, re2) with
-              (* DEDICATED RPFS REDUCTIONS *)
-              | NOp (LstCat, _), LVar y when rpfs && prefix_catch pfs re1 y ->
-                  Eq (UnOp (LstLen, re1), UnOp (LstLen, re2))
```
N1ark, line 2673: Removed this, as it incorrectly converted list equalities to list length equalities, and I'm not sure how that makes any sense (might be related to the fact I removed the `rpfs` parameter)

GillianCore/engine/FOLogic/Reduction.ml
```diff
@@ -1697,410 +1324,905 @@ and reduce_lexpr_loop
...
-                        (fun e ->
-                          Formula.equal e (Eq (flel, fler))
-                          || Formula.equal e (Eq (fler, flel)))
-                        pfs
-                    then Lit (Bool true)
-                    else if
-                      PFS.mem pfs (Not (Eq (flel, fler)))
-                      || PFS.mem pfs (Not (Eq (fler, flel)))
-                    then Lit (Bool false)
-                    else
```
N1ark, line 1771: Removed this, because checking the PFS for the equality would always just find the currently reduced expression, so it would always simplify to true.
This didn't happen before since there was a split between Expr/Formula and this case only applied to Expr.

GillianCore/engine/general_semantics/subst.ml
```diff
@@ -324,6 +323,8 @@ module Make (Val : Val.S) : S with type vt = Val.t = struct
                        generating fresh: %s"
                       x lvar));
               Expr.LVar lvar)
+
+        (* Need to handle visit_ForAll ?? *)
```
NatKarmios, line 327: And this

NatKarmios (approved):
A bug'un! Looks fine overall, nice stuff.
Happy to merge once comments are resolved :)

GillianCore/GIL_Syntax/Expr.ml
```diff
@@ -466,6 +518,78 @@ let loc_from_loc_name (loc_name : string) : t =
...
+      let+ inner, inner_neg = f inner in
+      let pos = Exists (bt, inner) in
+      let neg = ForAll (bt, inner_neg) in
+      (BinOp (pos, Equal, true_), neg)
+  | ForAll (bt, e) ->
+      let+ inner, inner_neg = f e in
+      let pos = ForAll (bt, inner) in
+      let neg = Exists (bt, inner_neg) in
+      (pos, BinOp (neg, Equal, true_))
+  | _ -> None
```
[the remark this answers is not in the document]
N1ark: Ah right! Yes it happens and sadly it happens in like, two Collections-C tests (iirc), literally nowhere else, but so I felt it wasn't worth re-factoring. In particular because when taking a look at the [Gillian-C profile for AWS](https://discord.com/channels/1028990778913067039/1029023172370108426/1313882400333234269) this function (or what it was called, `Formula.lift_logic_expr`) doesn't even show up meaning it took <1% of its parent's time, so no point bothering too much with it. I'll remove the `TODO`

GillianCore/engine/general_semantics/subst.ml
N1ark: Ah! For this @giltho might be able to help. The original subst for `Formula.ForAll` does some special handling around bound variables. However it doesn't do it for `Expr.Exists`, or `Expr.EForAll`, so I'm not sure how it makes any sense. Also this was removed in the PR and all tests still pass ?? So not sure if it should be re-added.
https://github.com/GillianPlatform/Gillian/blob/467a34846c6c6523ccb775b0b87ca6fa78e9bc07/GillianCore/engine/general_semantics/subst.ml#L348-L379

GillianCore/GIL_Syntax/Expr.ml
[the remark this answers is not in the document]
giltho: Which case happens? The none?
Honestly I would only have a function to negate and if really needed one can just call Typing.ml to type something to bool?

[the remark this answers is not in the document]
giltho: Even if it's not computation intensive, it's an indirection that adds complexity to the code for no reason

GillianCore/engine/general_semantics/subst.ml
giltho: Substituting definitely requires special handling for forall and exist to be sound. It wasn't there?

giltho: Ah I thought I did remember implementing those! I even fixed the visit_ForAll that was all over the place, I had a conversation with Petar about t.
What I didn't realise is that there are... two kinds of substitutions...

It's correctly implemented in eSubst.ml, copy from there. I'll fix it in my branch too cause it might case some nightmares in Rust

giltho: test

GillianCore/engine/FOLogic/Reduction.ml
giltho: Yeah remove that one, but again: this is only extremely slow because of time spent typing expressions ;)

GillianCore/engine/FOLogic/Reduction.ml
giltho: Nice

GillianCore/engine/FOLogic/Reduction.ml
N1ark: I imagined 🫠

GillianCore/GIL_Syntax/Expr.ml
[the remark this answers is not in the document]
N1ark: Ohhhhh this is so much nicer -- done in 137c679

GillianCore/engine/general_semantics/subst.ml
N1ark: Fixed in 57164a0

GillianCore/engine/FOLogic/Reduction.ml
N1ark: Mentioned in 4915efc

NatKarmios (approved):

PR #329, opened 2025-01-13
Fix branch max on calls during ACT

2over12:
Fixes from in progress [LLVM support](https://github.com/trail-of-forks/Gillian/pull/1)

When executing with a spec, regardless of how many specs are matched successfully etc, the call is considered branching which causes unexpected requirements on call factors. This causes unexpected unrolling requirements on examples like: https://github.com/trail-of-forks/Gillian/blob/9db721dc9cb70c3914dd28d1ef383f6b99c98a63/Gillian-LLVM/examples/wpst/mem_test.gil#L15 where 3 is required to handle all of the calls.

The intended fix here is to just add the number of matched specs exceeding 1 call

### DIFF
--- a/GillianCore/engine/general_semantics/general/g_interpreter.ml
+++ b/GillianCore/engine/general_semantics/general/g_interpreter.ml
@@ -857,11 +857,11 @@ struct
                       | Error x -> Right (Exec_err.EState x))
                     ret
                 in
-                let b_counter, has_branched =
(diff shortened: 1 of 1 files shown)


NatKarmios:
And thanks again :D Will hopefully review soon.

NatKarmios:
Closing in favour of #375
Thanks :)

PR #330, opened 2025-01-13
Fix for callgraphs used in ACT when using already compiled mode

2over12:
Fixes from in progress https://github.com/trail-of-forks/Gillian/pull/1

When executing ACT using -a the call graph is not populated. The lack of call graph causes an arbitrary ordering of functions preventing specs from being derived. For example in:   https://github.com/trail-of-forks/Gillian/blob/9db721dc9cb70c3914dd28d1ef383f6b99c98a63/Gillian-LLVM/examples/wpst/mem_test.gil test_add would get executed after split.

The intended fix here is during parsing to annotate callees.

### DIFF
--- a/GillianCore/gil_parser/gil_parsing.ml
+++ b/GillianCore/gil_parser/gil_parsing.ml
@@ -57,6 +57,13 @@ module Make (Annot : Annot.S) = struct
     in
     { labeled_prog; init_data }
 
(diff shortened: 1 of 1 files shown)


NatKarmios:
Thanks for this :) We'll review this as soon as we can.

NatKarmios:
Closing in favour of #375
Thanks :)

Issue #281, opened 2024-04-02
Rework debugger UI


NatKarmios:
Having explicit ExecMap and MatchMap were Gillian-specific choices that should be worked on more.

Changes should include:
- [ ] Allow sending multiple execution maps
- [ ] Unify ExecMap and MatchMap to a single packaged structure
- [ ] Allow sending only the updated parts of exec maps (to improve debugger performance)
- [ ] Allow any arbitrary submap to be isolated in a new tab

Issue #269, opened 2023-11-03
Compound branching


NatKarmios:
It'd be a neat idea if, in the debugging interface, "compound branches" i.e. multiple branches between two commands, could be split out and represented separately

i.e.:
```
        +---------------+
        | if (x.y == 1) |
        +---------------+
              /   \
 (x != null) /     \ (x == null)
            /       \
           /     +------+
          o      | fail |
         / \     +------+
        /   \
(true) /     \ (false)
      /       \
  +-----+   +-----+
  | ... |   | ... |
  +-----+   +-----+
```

Might implement this as part of Kanillian stuff.

Issue #233, opened 2023-03-24
Make debugger interface accessible via webserver


NatKarmios:
We don't actually need the interface to be embedded inside VSCode.

This isn't a priority, but it'd certainly be nice to have.

Issue #194, opened 2022-12-05
Add breakpoint marker to ExecMap


NatKarmios:
It'd be nice to show which cmds have breakpoints in the ExecMap to make it more consistent with the default debugger interface.

PR #333, opened 2025-02-06
Transition debugger to SEDAP

### LINKED ISSUE #110: Make frontend code more consistent with the rest of the repo
At the very least, this entails making types draw from common code. This could also involve rewriting some or all of the frontend in Caml.

Currently attempting to use ReScript to compile Caml for the frontend, which is going well except for [this speed bump](https://github.com/rescript-association/genType/issues/616).

We could potentially use [Melange](https://jchavarri.github.io/melange-docs/) instead of ReScript to have more confidence in continued OCaml support.

NatKarmios:
It's SEDAP time!

Main changes include:
- Use [`ocaml-sedap`](https://github.com/GillianPlatform/ocaml-sedap) instead of `ocaml-dap`
- Reconfigure debugger to send SEDAP information; in particular:
  - Only send changed nodes on update, instead of sending *everything* every update
  - Match maps are sent alongside exec maps in the same format
- Remove `debug-ui`; the Gillian debugger extension now lives in [the SEDAP-JS repo](https://github.com/GillianPlatform/sedap-js/tree/main/examples/gillian-debugging)

Due to the debug extension now being out of scope of the repo, this PR resolves #110, #173, #177, #181, #194, #233, #269, #281, #286


Gillian.code-workspace
```diff
@@ -0,0 +1,49 @@
+{
```
N1ark, line 1: Just checking; this file being included is intended?

GillianCore/logging/log_database.ml
```diff
@@ -189,3 +189,38 @@ let get_children_of roots_only id =
...
+  in
+  Sqlite3.finalize stmt |> check_result_code db ~log:("finalize: " ^ msg);
+  report
+
+let get_parent_of id =
+  let db = get_db () in
+  let stmt =
+    Sqlite3.prepare db
+      "SELECT id, type, content FROM report WHERE id = (SELECT parent FROM \
+       report WHERE id = ?);"
```
N1ark, line 211: Could we not prepare the query on module instantiation and then re-use the same query rather than always re-compiling it? / Or use some `stmt option ref`? No clue how expensive/how often this fn is run so maybe its pointless

N1ark (approved):
looks good to me :) there's some repetition between the lifters for C2 and WISL but I imagine we can't do much about that bc of different types?

NatKarmios:
> looks good to me :) there's some repetition between the lifters for C2 and WISL but I imagine we can't do much about that bc of different types?

Yes, the lifters are extremely similar, but factoring things out there felt like over-optimising. This might change if/when we have a JS lifter since that might be similar too.

N1ark (dismissed):
so cool!

Gillian.code-workspace
NatKarmios: Eh, it's probably a bit too opinionated to include in the repo.
Removed in cf6f27ab4db6ab1dd9305c3762098c30a72a61e8

GillianCore/logging/log_database.ml
NatKarmios: Not sure how expensive, but might as well :-)
Done in b8d695478adce1f7e6b821c356708b869b9ff727

PR #337, opened 2025-03-12
Some LSP tweaks for lab testing

NatKarmios:
- Change `Config.verify_only_some_of_the_things` to `Config.things_to_verify`, which can be one of
  - `All` - everything; equivalent to `false` before
  - `Specific` - specific procs and lemmas; equivalent to `true` before
  - `ProcsOnly` - verify all procs, but no lemmas
  - `LemmasOnly` - verify all lemmas, but no procs
- Add a `--procs-only` switch to verification debgger & verification language server to make use of this
- Track the "last known location" in `g_interpreter`; errors will now be attributed to the most recent location *that isn't in an internal proc*

### DIFF
--- a/GillianCore/command_line/verification_console.ml
+++ b/GillianCore/command_line/verification_console.ml
@@ -18,6 +18,10 @@ module Make
     let doc = "Do not verify the proofs of lemmas." in
     Arg.(value & flag & info [ "no-lemma-proof" ] ~doc)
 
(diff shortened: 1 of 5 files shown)


N1ark (approved):
lgtm! I am not too keen on `things_to_verify`, would have probably called it `verification_target`, but it's ok as is

PR #339, opened 2025-04-14
JS: Add check for division by zero

### LINKED ISSUE #338: Global predicates
Hello,

This is not an issue, more like a question. I want to add global predicates to Gillian-JS

So Gillian-JS will automatically check for runtime errors, such as:

    Division by zero (b === 0).

    Type errors (if a or b are non-numeric).

in the code below

```
function divide(a, b) {
  return a / b;
}

var result = divide(10, 0);
```
So user do not have to write any annotations (assumes, asserts).

What is the best way/path to do it?

klausnat:
Tweak the JS compiler to check for zero when performing `Div` or `Mod`.
Resolves #338.


NatKarmios (approved):

giltho:
Funily enough, this is invalid 
Gillian-JS doesn't check for division by zero because... well it's not a bug, in JS it just returns infinity.

PR #341, opened 2025-04-30
Fixes for WISL, debugging, loop invariants

NatKarmios:
### Main changes
  - Fix debugging for WISL invariants
  - Add `proc_hidden` and `proc_display_name` fields
    - These allow some customisation for how a proc's root node is displayed in the debugger
    - Here, it's used to make WISL loop bodies look nicer
  - Add type assertions to WISL
  - Improved handling of SMT failures; [see below for a more detailed explanation](https://github.com/GillianPlatform/Gillian/pull/341#issuecomment-2855319697)
  - Changed `SLL_iterative.wisl` to use the iterative functions from the old lab exercises
    - This file is now checked as part of the "quick WISL checks" in CI
  - Handle an edge case of preprocessing, where `l+ (x (rev y))` was causing failure
### Minor changes
  - Preprocessing errors are now considered analysis failures, rather than compilation errors
  - When failing to create an MP for a post-condition, Gillian now fails outright instead of just warning, as this creates more confusing errors later down the line
  - Changed CI to not restore stale caches; this was causing issues with updating dependencies
  - Bumped the required version of `linol` to ensure their vendored version of `lsp` is available


NatKarmios:
# A note on SMT error handling

To recap, we had an issue where some SMT queries cause Z3 to stack-overflow and segfault. When this happens, the next time that Gillian tries to interact with Z3 (the next time Gillian writes to the pipe, to be more precise), a `SIGPIPE` is raised. We now handle this properly with an exception instead of instantly crashing.

However, we then have an issue of only "noticing" that Z3 has died the next time we try to use it, which may be during an otherwise non-problematic query. To fix this, each command to Z3 is followed by a "heartbeat", where we run `(echo "gillian")` and check the response to make sure the solver is still alive.

N1ark (approved):
lgtm !

PR #351, opened 2025-05-26
SSFT lab changes

NatKarmios:
Many changes, including:
- The lifter now decides how to print the state when debugging
  - WISL's lifter now overhauls the state display
- L-var names have changed
  - These are intended to look more consistent with other lvars, like so:
    - `_lvar_5` -> `#lvar_5`
    - `_$l_0` -> `#loc_0`
  - [ ] This is causing breakages in C and JS tests! Need to fix this.
- Add usage logs; these can track:
  - The result of an analysis performed by the LSP
  - Start, stop, and various interactions when debugging
- WISL compiler changes:
  - Make trailing semicolons optional
  - Add in-file config statement to ignore particular functions and lemmas
- Various UI improvements in line with SEDAP updates
- Show errors more clearly when debugging


NatKarmios (commented):
I've given it a look over 2+ months after writing most of this, surely that's close enough to a review 😛
Certainly since nobody else is going to review this 😂

PR #352, opened 2025-07-09
Remove all notions of variants from the Gillian engine

giltho:
It's currently unused and should be done better anyway


N1ark (approved):
nice

PR #356, opened 2025-09-12
OCaml 5.3

NatKarmios:
- Update OCaml to 5.3
  - Use new effect handler syntax
- Update `ocamlformat` to 0.27.0 (required for compatibility)
  - Many files changed because `ocamlformat` formats doc comments now
- Use bash to install git hooks instead of an OCaml script


N1ark:
divine intervention

N1ark:
🤣🤣🤣🤣

PR #359, opened 2025-10-03
Add `wislf`, parallel calls, remove some errors

N1ark:
- Add `wislf`, a fractional version of WISL 
- Add `Par` calls to GIL, for parallel calls
- Fix some errors: in many places, vanishing was considered an error, which it isn't: e.g. in substitution, or in consume. I think -- at least these errors came up when I initially removed `matching`, despite the specs being sound.
- `matching` has not been removed here, for performance issues in the C AWS case study, but the commit (ed2222a9134fc537e9e86b388b1ffbff7cbb163d) still exists so one can potentially reapply the changes in the future if needed


GillianCore/monadic/MonadicSMemory.ml
```diff
@@ -97,13 +97,11 @@ module Lift (MSM : S) :
       (t * Expr.Set.t * (string * Type.t) list) list =
     let process = substitution_in_place subst mem in
     let curr_pc = Pc.make ~matching:false ~pfs ~gamma () in
-    match Delayed.resolve ~curr_pc process with
-    | [] -> failwith "substitution killed every branch, that cannot happen"
-    | leeloo_dallas_multibranch ->
-        List.map
-          (fun (bch : t Branch.t) ->
-            (bch.value, bch.pc.learned, bch.pc.learned_types))
-          leeloo_dallas_multibranch
```
NatKarmios: leeloo-dallas-fifth-element

PR #377, opened 2026-06-28
Only activate division check based on a CLI flag

giltho:
#339 introduces a regression because division by zero is valid in JS. But sometimes, users may actually want to check for this. We hide the feature behind a  `--forbid-div-by-zero` flag


giltho:
If you say gillian-c has been spuriously failing already, let's merge this? @NatKarmios

PR #391, opened 2026-07-09
Predicates as a combinator, the final stretch

giltho:
Based on #390 and #388 
Sorry i'm giving up on spliiting into PRs it's taking me too much time. I'm still splitting into adequate commits.

The predicate combinator is still, effectively, in the making, but I've had to do the following refactors so far (in addition to the big refactors in #380 #381 #382 #385 #386 #387 #388 and #390):
- Make everything "MonadicSMemory" (since the predicate combinator is, it needs its inputs to be). This means rewriting JS to use immutable memory and monadic from scratch.
- make wisl immutable. I'm pretty sure it's unsound otherwise. In fact, I removed all notions of immutability in memory models.
- remove all dead code after this cleanup


giltho:
I honestly don't know how you go at reviewing this
Well, commit-by-commit for sure, but this is massive....

